PR #34, opened 2021-03-08
Added empty audio recording activity

Xelowak:
Added empty audio recording activity so that we can link it.
Added user permission to use the microphone.


Lollipoke:
Code coverage test

PR #39, opened 2021-03-11
Splash screen

cmsigrist:
Implements a SplashScreenActivity that replaces the MainActivity as the LAUNCHER Activity.
Also changed the default Android logo to the Blindly logo.

- [x] Unit testing still need to be done (though what can be tested in an Activity containing only an ImageView and firing no Intent?)


cgntnr (changes requested):
Some resources seem to be unused so they can be taken out, also contentDescription should be added for the logo for accessibility issues.

app/src/main/AndroidManifest.xml
```diff
@@ -16,13 +16,20 @@
...
-
+              
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        <activity android:name=".MainActivity">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+              
+                <category android:name="android.intent.category.DEFAULT" />
```
Xelowak, line 30: MainActivity used to have android.intent.category.LAUNCHER before (instead of DEFAULT now). Might be an issue. See lines 19, 23, 26 and 30.

Xelowak (approved):
Seems to be ok

cgntnr (approved):
Some warnings are still there (many of those from my previous pull request), I will correct them for the next merging. Otherwise all good.

PR #40, opened 2021-03-11
draft for location 

r-ismer:
- [x] maps activity
- [x] location permission
- [x] get the location 
 
the maps activity dispay well on the location entered but the permissions and the location are not working for now


r-ismer:
I don't test the main activity because this one will not call the other activities. This is why the test coverage is below 80%.

cmsigrist (approved):
Maybe use @string resource instead of harcoded strings for the text of the buttons in the activity_location_permission. Otherwise it looks good to me!

app/src/main/java/ch/epfl/sdp/blindly/LocationPermissionActivity.kt
```diff
@@ -0,0 +1,32 @@
...
+        setContentView(R.layout.activity_location_permission)
+
+        val button: Button = findViewById(R.id.button)
+
+        button.setOnClickListener{
+            var permitted = ActivityCompat.checkSelfPermission(this, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED
+            while(!permitted) {
+                ActivityCompat.requestPermissions(this, arrayOf(Manifest.permission.ACCESS_FINE_LOCATION), FINE_LOCATION_PERMISSION_CODE)
+                permitted = ActivityCompat.checkSelfPermission(this, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED
+            }
```
Xelowak, line 26: Is it really relevant to ask for permissions on a button click ?

app/src/main/java/ch/epfl/sdp/blindly/location/AndroidLocationService.kt
```diff
@@ -0,0 +1,56 @@
...
+                } else {
+                    //get location from Network
+                    locationManager.requestLocationUpdates(LocationManager.NETWORK_PROVIDER, MIN_TIME_FOR_UPDATE,  MIN_DISTANCE_FOR_UPDTAE, this)
+                    locationManager.getLastKnownLocation((LocationManager.NETWORK_PROVIDER))
+                }
+
+            }
+
+        } catch(e: SecurityException) {
+            throw e
```
Xelowak, line 48: Might be useful to do something with this SecurityException instead of throwing it

app/src/main/java/ch/epfl/sdp/blindly/location/LocationService.kt
```diff
@@ -0,0 +1,9 @@
+package ch.epfl.sdp.blindly.location
+
+import android.location.Location
+import android.location.LocationListener
+
+interface LocationService: LocationListener {
+
+    fun getCurrentLocation(): Location?
+}
```
Xelowak, line 9: Is this interface relevant ?

Xelowak (approved):
It's ok. Be careful with hardcoded strings though

app/src/main/java/ch/epfl/sdp/blindly/MapsActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+     * it inside the SupportMapFragment. This method will only be triggered once the user has
+     * installed Google Play services and returned to the app.
+     */
+    override fun onMapReady(googleMap: GoogleMap) {
+        mMap = googleMap
+
+        val loc: LatLng = if(location != null) {
+            LatLng(location!!.latitude, location!!.longitude)
+        } else {
+            LatLng(46.5, 6.6)
```
Lollipoke, line 50: What is LatLng(46.5, 6.6) referring to ? The marker near Sydney ? Might want to add a comment or a constant

Lollipoke (approved):

app/src/main/java/ch/epfl/sdp/blindly/LocationPermissionActivity.kt
r-ismer: Yes because they have to read that the location is necessary to use the app

app/src/main/java/ch/epfl/sdp/blindly/location/LocationService.kt
r-ismer: If we want to add another location service it would be easier since it is more modular

app/src/main/java/ch/epfl/sdp/blindly/MapsActivity.kt
r-ismer: The comment above explains this but it is not Sydney but Lausanne instead

PR #61, opened 2021-03-16
Passions Screen

cgntnr:
Passions Screen is added to the Profile Preferences section.


app/src/main/java/ch/epfl/sdp/blindly/Profile6.kt
```diff
@@ -1,11 +1,18 @@
...
 import android.os.Bundle
+import android.view.View
 
 class Profile6 : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.set_profile_6)
     }
+    fun start_profile_7(view: View) {
+        // Do something in response to button
```
koko-ng, line 14: I think this comment could be deleted.

koko-ng (approved):
Alright for me.

app/src/main/java/ch/epfl/sdp/blindly/Profile6.kt
cgntnr: Yes you are right, I will delete this in the next PR.

Lollipoke (approved):
Seems good

PR #62, opened 2021-03-17
Profile page

cmsigrist:
Add a Profile page with a Settings page


app/src/main/java/ch/epfl/sdp/blindly/SplashScreen.kt
```diff
@@ -4,22 +4,28 @@ import android.content.Intent
...
         supportActionBar?.hide()
-        window.decorView.apply {
-            systemUiVisibility = View.SYSTEM_UI_FLAG_HIDE_NAVIGATION or View.SYSTEM_UI_FLAG_FULLSCREEN
-            }
 
+        val heart = findViewById<ImageView>(R.id.splashscreen_heart)
+        val beating = AnimationUtils.loadAnimation(this, R.anim.beating_heart)
+
+        // Use bounce interpolator with amplitude 0.2 and frequency 20
+        val interpolator = BounceInterpolator(0.3, 20.0)
```
r-ismer, line 25: not the same constants

app/src/main/java/ch/epfl/sdp/blindly/settings/Settings.kt
```diff
@@ -0,0 +1,92 @@
...
+import android.view.View
+import android.widget.SeekBar
+import android.widget.TextView
+import androidx.appcompat.app.AppCompatActivity
+import ch.epfl.sdp.blindly.R
+
+
+const val EXTRA_LOCATION = "user_location"
+const val EXTRA_SHOW_ME = "user_show_me"
+//const val REQUEST_LOCATION = 1
```
cgntnr, line 14: Is this comment still needed, also are these used outside, maybe they could be private ?

cgntnr (approved):
Added a note on a comment and making vals private when possible otherwise seems okay to me.

r-ismer (approved):
Seems very clean and clear to me, I approve!!

app/src/main/java/ch/epfl/sdp/blindly/settings/Settings.kt
cmsigrist: They're used by the children Activities, and the commented one will be used next week

cgntnr: All good, thanks!

PR #64, opened 2021-03-18
User class

Lollipoke:
- [x] User Class

- [x] User Class Tests


app/src/main/java/ch/epfl/sdp/blindly/User.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Sets the different attributes with the values of the User
+     * @param birthDate the User birthday
+     * @param gender the User gender
+     * @param sexualOrientation the User sexual orientation
+     * @param desiredGender the genders the User wants to be match with
+     * @param passions the passions of the User
+     */
+    fun fillProfile(birthDate: String, gender:String, sexualOrientation: Array<String>,
```
koko-ng, line 59: Not sure about this one, looks kind of like half a constructor. I think either a full constructor or setters would make more sense.

app/src/main/java/ch/epfl/sdp/blindly/User.kt
```diff
@@ -0,0 +1,72 @@
...
+ * Represents a user of the app.
+ * Contains login informations and preferences
+ * @param name the User name
+ * @param email the User email
+ */
+class User(private val name: String, private val email: String){
+    /**
+     * User birthdate
+     */
+    var birthDate: String = DateFormat.getDateInstance().format(Calendar.getInstance().time) //Today
```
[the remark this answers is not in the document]
r-ismer: agree with this comment

app/src/main/java/ch/epfl/sdp/blindly/User.kt
r-ismer: could be usefull when retrieving from the database and create a profile in one line (or even for testing the profile class)

r-ismer (approved):
change some java style to kotlin but globally very good

app/src/main/java/ch/epfl/sdp/blindly/User.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * User gender
+     */
+    var gender: String = ""
+
+    /**
+     * User sexual orientation, up to 3
```
Xelowak, line 37: Might be a good idea to create a constant number representing the max number of sexual orientation instead of writing it in a comment.

app/src/main/java/ch/epfl/sdp/blindly/User.kt
```diff
@@ -0,0 +1,72 @@
...
+     */
+    var sexualOrientation: Array<String> = emptyArray()
+
+    /**
+     * Gender user wants to be matched with
+     */
+    var desiredGender: DesiredGender = DesiredGender.MEN
+
+    /**
+     * User passions, up to 5
```
Xelowak, line 47: Same comment as for sexual orientations (create a constant).

app/src/main/java/ch/epfl/sdp/blindly/User.kt
```diff
@@ -0,0 +1,72 @@
...
+     */
+    var passions: Array<Passions> = emptyArray()
```
Xelowak, line 49: Are those comments really relevant ? It seems logical for me and doesn't need to be specified.
Moreover, shouldn't all these variables be private ?

app/src/main/java/ch/epfl/sdp/blindly/User.kt
```diff
@@ -0,0 +1,72 @@
...
+     * @param gender the User gender
+     * @param sexualOrientation the User sexual orientation
+     * @param desiredGender the genders the User wants to be match with
+     * @param passions the passions of the User
+     */
+    fun fillProfile(birthDate: String, gender:String, sexualOrientation: Array<String>,
+                    desiredGender: DesiredGender, passions: Array<Passions>){
+        //Argument check
+        require(sexualOrientation.size <= 3)
+        require(passions.size <= 5)
```
Xelowak, line 63: Linked to other comments (magic numbers => create a private constant)

app/src/main/java/ch/epfl/sdp/blindly/User.kt
```diff
@@ -0,0 +1,72 @@
...
+    var passions: Array<Passions> = emptyArray()
+
+    /**
+     * Sets the different attributes with the values of the User
+     * @param birthDate the User birthday
+     * @param gender the User gender
+     * @param sexualOrientation the User sexual orientation
+     * @param desiredGender the genders the User wants to be match with
+     * @param passions the passions of the User
+     */
```
Xelowak, line 58: Again, I don't think this function is worth documenting.

app/src/main/java/ch/epfl/sdp/blindly/User.kt
Xelowak: Since each attribute will be set from another activity (we have different activities for birth date, age, name, etc...), I don't think it is useful to create this constructor-like function that we don't need in Kotlin.

Xelowak (changes requested):
It may be better to use Hamcrest's `assertThat ` in the test class for clarity and consistency.

Lollipoke:
User class design has been modified in another branch to fit the database

PR #65, opened 2021-03-19
integrate UserHelper with the rest

koko-ng:
Just some code to integrate the authentication process into our current app.


app/src/main/java/ch/epfl/sdp/blindly/BlindlyApplication.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import android.app.Application
+import android.content.res.Configuration
+import dagger.hilt.android.HiltAndroidApp
+
+// Custom app to let hilt hook into
+@HiltAndroidApp
+class BlindlyApplication : Application() {
+
+}
```
Xelowak, line 11: Why is this empty class relevant ?

app/src/main/java/ch/epfl/sdp/blindly/utils/UserHelper.kt
```diff
@@ -6,109 +6,123 @@ import android.content.Intent
...
+
+        return AuthUI.getInstance()
+            .createSignInIntentBuilder()
+            .setAvailableProviders(providers)
+            //.setLogo(R.drawable.my_great_logo) // Set logo drawable
+            .setTheme(R.style.Theme_Blindly) // Set theme
+            /*.setTosAndPrivacyPolicyUrls(
+    "https://example.com/terms.html",
+    "https://example.com/privacy.html")*/
+            .build();
```
Xelowak, line 58: Might be a good idea to clean up these comments and keep only what is useful (or explain why it is in comments for now).

Xelowak (changes requested):
Might be useful to explain why all `@RunWith(AndroidJUnit4::class)` have been replaced by  `@HiltAndroidTest`, and how it works.  
Will we have to do that for each new test class that we will write in the future ?

In general, I think it is quite difficult to understand everything that has been done by reading your code. You should consider commenting it more since others (including me) don't know anything about these libraries and annotations that don't seem to be really instinctive (like Hilt).

app/src/main/java/ch/epfl/sdp/blindly/BlindlyApplication.kt
koko-ng: Just read the comment above ;)

app/src/main/java/ch/epfl/sdp/blindly/utils/UserHelper.kt
koko-ng: Well for the comments it is just have these codes available when we will need them, I think adding login by phone could  interset us. Then we will probably add ToS and policy one day.

I admit that I could already add the logo now.

koko-ng:
> Might be useful to explain why all `@RunWith(AndroidJUnit4::class)` have been replaced by `@HiltAndroidTest`, and how it works.
> Will we have to do that for each new test class that we will write in the future ?

Well I don't know what is more intuitive between RunWith and HiltAndroidTest, the first one literally tells that it runs a with Android JUnit while the second tells that it's an android test with hilt. I mean I don't think it brings a lot of difference, and commenting everything with "hook hilt into this test" looks quite redundant to me...

app/src/main/java/ch/epfl/sdp/blindly/BlindlyApplication.kt
cmsigrist: Is this to inject manual dependencies?
I'm not sure I understand how this works... Couldn't we do without this empty class?

cmsigrist (approved):

Xelowak:
> > Might be useful to explain why all `@RunWith(AndroidJUnit4::class)` have been replaced by `@HiltAndroidTest`, and how it works.
> > Will we have to do that for each new test class that we will write in the future ?
> 
> Well I don't know what is more intuitive between RunWith and HiltAndroidTest, the first one literally tells that it runs a with Android JUnit while the second tells that it's an android test with hilt. I mean I don't think it brings a lot of difference, and commenting everything with "hook hilt into this test" looks quite redundant to me...

I didn't mean that you had to comment all annotations. However, you could tell us the changes you made (and why) in the description of the pull resquest.

Xelowak (approved):
We talked about these changes in our meeting, it's ok for me.

PR #85, opened 2021-03-23
Input checks

cgntnr:
* Input checks are completed for profile preferences pages.
* Hints are added for usability for text inputs.
* For Profile7, since we haven't done the linking part; button sends back to MainActivity in case of success. I've added TODOs and comments so that we correct these once linking activities are done.


Lollipoke (approved):
Really nothing to say about your code 👍

cgntnr:
Unfortunately I had to comment out two tests for Passions page however Cirrus couldn't seem to perform clicks on chips on that page. The tests worked fine locally.

app/src/main/res/layout/set_profile_4_2.xml
```diff
@@ -25,6 +25,7 @@
         android:layout_height="50dp"
         android:contentDescription="@string/please_specify_gender"
         android:ems="10"
+        android:hint="@string/more_hint"
```
r-ismer, line 28: does the hint get covered by the contentDescription ? if it is the case remove one of the two

r-ismer (approved):
very good work ! looks very good to me. At some point we will have to check if the strings are actually used so that cirrus don't display warnings

app/src/main/res/layout/set_profile_4_2.xml
cgntnr: hint is the one that appears greyed out before you start to type something. contentDescription was for Accessibility if I'm not mistaken.

r-ismer (commented):
everything is clear, looks good to me

PR #86, opened 2021-03-24
Recording UI, part 1

Xelowak:
I added a Profile8 after the passions page. This activity explains that the user will need to record an audio file. After pressing continue, the recordingActivity is launched.

### DIFF
--- a/app/src/main/res/layout/set_profile_8.xml
+++ b/app/src/main/res/layout/set_profile_8.xml
@@ -51,16 +51,15 @@
         android:layout_width="196dp"
         android:layout_height="wrap_content"
         android:layout_marginStart="35dp"
-        android:layout_marginTop="57dp"
+        android:layout_marginTop="81dp"
         android:layout_marginEnd="180dp"
-        android:layout_marginBottom="278dp"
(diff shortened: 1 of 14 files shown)


r-ismer (approved):
LGTM !

Lollipoke (approved):
Seems good to me

app/src/main/res/layout/set_profile_8.xml
```diff
@@ -0,0 +1,65 @@
...
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintVertical_bias="0.167" />
+
+    <ProgressBar
+        android:id="@+id/progressBar_p8"
+        style="?android:attr/progressBarStyleHorizontal"
+        android:layout_width="413dp"
+        android:layout_height="62dp"
+        android:progress="89"
+        app:layout_constraintBottom_toTopOf="@+id/text_p2"
```
cgntnr: This seems to come from a copy-paste of another class, should be removed from constraints.

cgntnr: Seems to be fixed! All good.

PR #87, opened 2021-03-24
integrated maps

r-ismer:
integrated the permisission for the location after the presentation screen
integrated the map into the settings->location


app/src/main/java/ch/epfl/sdp/blindly/profile/Profile1.kt
```diff
@@ -13,7 +14,7 @@ class Profile1 : AppCompatActivity() {
     }
 
     fun startProfile2(view: View) {
```
cgntnr, line 16: Function name can be updated since it doesn't actually start Profile 2 right now, but I don't think we should start with asking Permissions right away in the profile. Similar apps seem to do it at the end of profile preferences.

app/src/main/java/ch/epfl/sdp/blindly/settings/SettingsLocation.kt
```diff
@@ -1,21 +1,71 @@
...
+        mapView.onCreate(savedInstanceState)
+        mapView.getMapAsync(this)
+
+    }
+
+    override fun onMapReady(googleMap: GoogleMap) {
+        val loc: LatLng = if (location != null) {
+            LatLng(location!!.latitude, location!!.longitude)
+        } else {
+            LatLng(46.5, 6.6)
```
cgntnr, line 45: What are these latitude and longitude values, can they be constants ?

cgntnr (changes requested):
I've added some comments I believe we shouldn't be asking permissions right at the beginning. Maybe this can be changed and merged after #85 .

app/src/main/java/ch/epfl/sdp/blindly/settings/SettingsLocation.kt
r-ismer: I agree they could have been contants (it's the latitude and longitud of lausanne)

app/src/main/java/ch/epfl/sdp/blindly/profile/Profile1.kt
r-ismer: We discussed it on whatsapp if we all agree to place it at the end it can be easily done

cgntnr: Exactly!

cgntnr (approved):

app/src/main/java/ch/epfl/sdp/blindly/profile/Profile1.kt
Xelowak, line 16: The function should be renamed, because she no longer fires Profile2

app/src/main/java/ch/epfl/sdp/blindly/settings/SettingsLocation.kt
```diff
@@ -1,21 +1,72 @@
...
+import com.google.android.gms.maps.model.MarkerOptions
 
-class SettingsLocation : AppCompatActivity() {
+
+class SettingsLocation : AppCompatActivity(), OnMapReadyCallback {
+
+    private lateinit var mapView: MapView
+    private var location: Location? = null
+    private lateinit var locSer: LocationService
+    private val LAUSANNE = LatLng(46.5, 6.6)
```
Xelowak, line 20: Wouldn't it be better to make a it a constant outside of the class ?

Xelowak (approved):
Seems good to me. Maybe add some contentDescription on images in activity_settings_location.xml as Cirrus warns us.

Issue #96, opened 2021-03-26
Add a map activity where we can see other users on it.


Xelowak:
Related to #101

koko-ng:
Now we can get the users location from the database, we should add them to the map.

PR #112, opened 2021-04-11
add map with pins and current position

koko-ng:
#96


cgntnr:
I didn't notice this sooner but just wanted to make sure we all understand the same thing. The task says "Add a map activity where we can see other users on it" but it's actually "only the current user and its matched user" that should see each other with live location, since our purpose is that matched people can find each other when they meet in person. I thought it's clear because that was what we talked in the meetings. Just wanted to clear the confusion.

koko-ng:
Yes thanks for the precisions, it's welcome :) . Anyway for now it's just a map that can be fed with different points, however it will be useful for the future :)

app/src/androidTest/java/ch/epfl/sdp/blindly/PermissionTest.kt
```diff
@@ -40,12 +46,10 @@ class PermissionTest {
 
     @Test
     fun testLocationPermissionFiresProfileName() {
-        activityRule.scenario.onActivity { activity ->
```
cmsigrist, line 43: I don't know if you were able to fix this (and thus removed this part) but sometimes there's a system dialog that pops up atop of the activity, and so cirrus cannot click on the button underneath, which at least used to make the tests fail?

app/src/androidTest/java/ch/epfl/sdp/blindly/main_screen/MapPageTest.kt
```diff
@@ -0,0 +1,90 @@
...
+    @After
+    fun afterEach() {
+        release()
+    }
+
+    private suspend fun goToProfileFragment() {
+        // Wait for everything to settle to avoid race conditions
+        // Waiting synchronously isn't a problem since there are no other
+        // tasks running and we don't care about UI responsivity while testing
+        suspendCoroutine { cont: Continuation<Boolean> ->
```
cmsigrist, line 64: There seem to be an easier way to start the ProfilePageFragment, by using transactions (cf ProfilePageTest) rather than coroutines, but then again "don't fix it if it ain't broke"...

app/src/main/java/ch/epfl/sdp/blindly/UserMapActivity.kt
```diff
@@ -0,0 +1,127 @@
...
+
+        // Add markers from the intent
+        val points: Array<LatLng>? = intent.getParcelableArrayExtra(POINTS)?.map { p -> p as LatLng }?.toTypedArray()
+        if (points != null) {
+            for (point in points) {
+                map!!.addMarker(MarkerOptions().position(point).title("Marker in Sydney2"))
+            }
+        }
+
+        // TODO move camera to user settings position
```
cmsigrist, line 63: You marked it as a TODO, but do we actually need the camera for the map?

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MapFragment.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.fragment.app.Fragment
+import ch.epfl.sdp.blindly.R
+import ch.epfl.sdp.blindly.UserMapActivity
+import com.google.android.gms.maps.model.LatLng
+
+
+class MapFragment : Fragment() {
+
+    companion object {
+        private const val ARG_COUNT = "messageArgs"
```
cmsigrist, line 18: You should change the string inside the ARG_COUNT to "mapArgs" ;)

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MapFragment.kt
```diff
@@ -0,0 +1,49 @@
...
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?,
+                              savedInstanceState: Bundle?): View? {
+        val view = inflater.inflate(R.layout.fragment_map_page, container, false)
+        val matchActivityButton = view.findViewById<Button>(R.id.open_map)
+        matchActivityButton.setOnClickListener { startUserMap() }
+        return view
+    }
+    private fun startUserMap() {
+        val intent = Intent(activity, UserMapActivity::class.java)
+        val points = arrayOf(LatLng(-33.0, 151.0))
```
cmsigrist, line 45: And also maybe remove these magic numbers and change them to a const DEFAULT_LOCATION_POINTS?

app/src/main/res/layout/fragment_map_page.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This linear layout is used to show elements
+     in vertical or in horizontal linear manner -->
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:gravity="center">
+
+    <!-- This TextView indicates new fragment is open -->
```
cmsigrist, line 10: Was there a TextView before? Maybe remove this comment since it doesn't seem to exist anymore

app/src/main/res/layout/user_map_activity.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context=".UserMapActivity">
+    <fragment xmlns:android="http://schemas.android.com/apk/res/android"
+        xmlns:tools="http://schemas.android.com/tools"
```
cmsigrist, line 9: And if you've got the time it would be great to clean up a bit the redundant namespace as Cirrus said and the unused resources (even though now that I look at it more into details most of them don't seem to come from your code x)  )

cmsigrist (approved):
Great work, thank you! 
I left some comments, there're just some minor things to modify :)

app/src/main/res/layout/user_map_activity.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
Lollipoke, line 2: Cirrus says we can use merge instead of FrameLayout, maybe we could check that ?

app/src/main/res/layout/user_map_activity.xml
```diff
@@ -0,0 +1,15 @@
...
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context=".UserMapActivity">
+    <fragment xmlns:android="http://schemas.android.com/apk/res/android"
```
Lollipoke, line 8: And here it says we could replace fragment with FragmentContainerView, and that the namespace declaration is redundant

app/src/main/res/layout/user_map_activity.xml
Lollipoke, line 9: this namespace declaration seems redundant too

Lollipoke: Ah, it was already said, mb haha

Lollipoke (approved):

Lollipoke (approved):
A few details, see my comments. Also I don't know if it's a really good idea to merge and loose 2.9% coverage, but I guess we can fix that later as we discussed on Whatsapp

app/src/androidTest/java/ch/epfl/sdp/blindly/PermissionTest.kt
koko-ng: Yes, I use the GrantPermissionRule in order to tell the test runner to automatically give permission, so it doesn't pop up.

app/src/androidTest/java/ch/epfl/sdp/blindly/main_screen/MapPageTest.kt
koko-ng: Well unfortunately in my tests in created a new fragment on top of the previous one and the buttons of the ProfilePage obfuscated my "open map" button, which then was not clickable. If you look at the emulator while running the ProfilePageTest you actually see that there is two different fragments on top of each other.

app/src/main/java/ch/epfl/sdp/blindly/UserMapActivity.kt
koko-ng: Oh yes, its working now indeed. I think I forgot it, thanks for the remark!

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MapFragment.kt
koko-ng: Indeed :)

app/src/main/java/ch/epfl/sdp/blindly/UserMapActivity.kt
koko-ng: By camera it is meant, having the user's position to the center of the map. However I don't get the user position for now, it is to be done when things are sent to firebase live.

PR #114, opened 2021-04-12
Facebook authentication

cgntnr:
Facebook authentication is added. I've tried and successfully added myself to the Firebase authenticated users. Note that it might not work out of the box for you (Give it a try after merge please). If I understood correctly from this [documentation](https://developers.facebook.com/docs/facebook-login/android/) described at step 6, every developer should generate development key unique to their development environment. Once generated I can add them via our registered Facebook Developers account.


app/build.gradle
```diff
@@ -58,6 +58,7 @@ dependencies {
     implementation 'com.google.firebase:firebase-firestore-ktx'
     implementation 'com.firebaseui:firebase-ui-auth:7.1.1'
     implementation 'com.google.android.gms:play-services-maps:17.0.0'
+    implementation 'com.facebook.android:facebook-android-sdk:8.+'
```
cmsigrist, line 61: Do you think there's a way to make it work without using the + ? Cirrus doesn't seem happy about it...

app/src/main/AndroidManifest.xml
```diff
@@ -68,6 +67,25 @@
...
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="@string/fb_login_protocol_scheme" />
+            </intent-filter>
+        </activity>
     </application>
-
+    <uses-permission android:name="android.permission.INTERNET" />
```
cmsigrist, line 90: I think (or at least Cirrus does) that you should move the <uses-permission> above the <application>

cgntnr: The [documentation] (https://developers.facebook.com/docs/facebook-login/android/) at step 4 specifically states to put this after application that's why I moved it. But further search on internet suggests to place it before as well even though there's no explanation whatsoever. I'll try it like that.

cmsigrist (approved):
Thanks for the modifications, it looks all good to me :)

Xelowak (approved):
LGTM !
Might it be possible to test your code, or is it too hard since Firebase is involved ?

cgntnr:
Unfortunately the same testing issues for Firebase authentication still applies here, I just added another method. The diff coverage is 0 but total-coverage seems untouched but maybe it's because it's way too little.

PR #115, opened 2021-04-13
Settings preference

Lollipoke:
Added a range slider for the age range selection in the settings
Modified the location radius to a slider for consitency with the age range
Added a delete account button for later
Added some documentation


app/src/main/java/ch/epfl/sdp/blindly/settings/Settings.kt
```diff
@@ -40,16 +41,30 @@ class Settings : AppCompatActivity() {
...
+            ageRangeSlider.values[0].toInt(),
+            ageRangeSlider.values[1].toInt()
+        )
+        ageRangeSlider.addOnChangeListener { _, _, _ ->
+            ageRangeText.text = getString(
+                R.string.selected_age_range,
+                ageRangeSlider.values[0].toInt(),
+                ageRangeSlider.values[1].toInt()
+            )
+        }
```
Xelowak, line 67: As codeclimate says, I think it might be better to create a private function to contain the lines 56 to 60 that are copy-pasted in lines 62 to 66.

app/src/main/res/layout/activity_settings.xml
```diff
@@ -68,10 +66,10 @@
                     android:id="@+id/email_address_text"
                     android:layout_width="wrap_content"
                     android:layout_height="match_parent"
+                    android:layout_marginRight="10dp"
```
Xelowak, line 69: Replace marginRight by marginEnd (Cirrus warning)

app/src/main/res/values/strings.xml
```diff
@@ -128,6 +130,8 @@
     <string name="check_mark_for_the_choice_everyone">Check mark for the choice Everyone.</string>
     <string name="lausanne_switzerland">Lausanne, Switzerland</string>
     <string name="progress_km">%1$dkm</string>
+    <string name="account_deleted">Your account has been deleted</string>
```
Xelowak, line 133: This string seems to be ununsed.

Xelowak (approved):
It seems nice ! Just correct some things I said in my comments. I saw a lot of spaces and tabs changes, so I would suggest you to format your code by using CTRL+ALT+L on each of the file you changed (for consistency).
Moreover, is it possible for you to search for ways to improve your code coverage ? I don't know if it's possible, but it may keep us at 80%.

app/src/main/java/ch/epfl/sdp/blindly/settings/Settings.kt
Lollipoke: Thing is it's already a single function, I only call getString so I thought using a private method was a bit overkill... But if you think it's better then I'll do it

app/src/androidTest/java/ch/epfl/sdp/blindly/settings/SettingsTest.kt
```diff
@@ -102,34 +103,61 @@ class SettingsTest {
...
         val radiusText = onView(withId(R.id.radius_text))
-        radiusText.check(ViewAssertions.matches(ViewMatchers.withText(TEST_RADIUS)))
+        radiusText.check(matches(withText(TEST_RADIUS)))
+        release()
+    }
+
+    @Test
+    fun movingTheRangeSliderChangesTheSelectedAgeRangeText() {
+        init()
+        onView(withId(R.id.age_range_slider)).perform(setRangeSliderValues(40, 60))
```
cmsigrist, line 158: Maybe you could replace the magic numbers 40 and 60 with constants

cmsigrist (approved):
Nice! And I see that you added doc for pre-existing functions:D

app/src/androidTest/java/ch/epfl/sdp/blindly/settings/SettingsTest.kt
Lollipoke: good idea :)

PR #133, opened 2021-04-18
Fix recording bugs + add tests

Lollipoke:
- [x] Recording playing that would not stop even when collapsed

- [x] Timers not updating when sliding the playbar

- [x] Clicking on the playbar when the recording is stopped should start it at the right place

- [x] Add tests for the audio recording activity and a test class for recyclerviews


Xelowak (approved):
Looks clean to me, nice work !
Thanks for implementing a class to test RecyclerView items ! It will surely be useful for other tests too.

Xelowak (approved):

cgntnr (approved):
Looks clean! We will make good use of the new tests. Thanks!

PR #134, opened 2021-04-21
Remove MainsActivity and reroute Login through SplashScreen

cmsigrist:
Since the rest of the app needs a user to be logged in, this forces the user to log in before doing anything else

- Remove the MainActivity
- Reroute the log in to be called from the SplashScreen instead
- Access to profile_setup only available when creating a new account
- Otherwise goes straight to the MainScreen
- If user did not log out from previous session will also go back to MainScreen
- Else will go to log in as usual


app/src/androidTest/java/ch/epfl/sdp/blindly/main_screen/ProfilePageTest.kt
```diff
@@ -55,23 +55,25 @@ class ProfilePageTest {
     }
 
 
+    //TODO the tests seem really unstable need to look into this,
+    // sometimes they pass other time they failed because of "Error performing single click on",
+    // or "Wanted to match 1 intents. Actually matched 0 intents."
```
r-ismer, line 60: This is very strange we sould look at it next week i think

app/src/main/java/ch/epfl/sdp/blindly/settings/Settings.kt
```diff
@@ -108,20 +115,15 @@ class Settings : AppCompatActivity() {
...
-                        Toast.LENGTH_LONG
-                    ).show()
+        AuthUI.getInstance()
+            .signOut(this)
+            .addOnCompleteListener { // user is now signed out
+                startActivity(Intent(this, SplashScreen::class.java))
+                finish()
+            }
+            .addOnFailureListener {
+                Log.e(TAG, "Error: Could not logout user.")
```
r-ismer, line 125: Maybe add a Toast to notify the user he couldn't logout

app/src/main/java/ch/epfl/sdp/blindly/user/UserHelper.kt
```diff
@@ -54,56 +52,65 @@ class UserHelper {
...
+     *
+     * @param user: the user to test the condition on
+     * @return true if the user just created an account, false otherwise
+     */
+    private fun isNewUser(user: FirebaseUser) : Boolean {
+        val metadata = user.metadata
+        return metadata?.creationTimestamp == metadata?.lastSignInTimestamp
+    }
+
+    fun delete(activity: Activity, onComplete: OnCompleteListener<Void>) {
```
r-ismer, line 95: Can't we use this function in the settings activity to delete the account ?

r-ismer (approved):
This is a good job ! This will improve the login a lot :) Some tests have to be checked and we might be able to fix them later

app/src/androidTest/java/ch/epfl/sdp/blindly/main_screen/ProfilePageTest.kt
cmsigrist: Yes I forgot to send the email with the stack trace to our assistants

app/src/main/java/ch/epfl/sdp/blindly/user/UserHelper.kt
cmsigrist: I don't know, I didn't write this part but probably (though there might be a bug, which is why it's not called yet)?

app/src/androidTest/java/ch/epfl/sdp/blindly/settings/SettingsTest.kt
```diff
@@ -207,9 +205,7 @@ class SettingsTest {
         init()
```
Lollipoke, line 205: You should update the name of the test like you did for the logout test

Lollipoke (approved):
Very great job ! You only forgot to change a test name, see my comment for details. Other than that your work looks good to me :)

PR #135, opened 2021-04-21
Matching algorithm

Xelowak:
Add a matching algorithm that works with users' accounts in the database.

For now, here are the criteria for an other user (O) to be presented to the current user (C):
1. O's gender corresponds to C's show me
2. At least one passion of C is in common with O
3. O is in the location range of C
4. O is in the age range of C, as specified in settings

Everything is then reverse checked (O and C are swapped)


app/src/main/java/ch/epfl/sdp/blindly/match/Profile.kt
```diff
@@ -2,6 +2,7 @@ package ch.epfl.sdp.blindly.match
 
 import android.os.Parcel
 import android.os.Parcelable
+import ch.epfl.sdp.blindly.user.User
```
cgntnr, line 5: Why was this import needed without any change in the file?

app/src/main/java/ch/epfl/sdp/blindly/match/UserListFilter.kt
```diff
@@ -0,0 +1,105 @@
...
+                    filteredList += user
+                }
+            }
+        }
+
+        return filteredList
+    }
+
+    /**
+     * Does the same checks than [MatchingAlgorithm.getPotentialMatchesFromDatabase], but from the
```
cgntnr, line 43: If they are doing the same checks wouldn't it be possible to merge these two and simply reverse the order of arguments passed. Though not sure if that would break anything, correct functioning comes first. I leave it up to you.

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
```diff
@@ -81,7 +83,7 @@ class User private constructor(
          * @param location the location of the User
          */
         fun setLocation(location: List<Double>) = apply {
-            if(location.size == SIZE_2)
+            if (location.size == SIZE_2)
```
cgntnr, line 86: I wrote that in a different PR but I still believe SIZE_2 defeats the purpose of making it a constant.

cgntnr (approved):
Great job! Added couple of comments, nothing big.

app/src/main/java/ch/epfl/sdp/blindly/match/Profile.kt
Xelowak: It was when I tried to implement a function in this class. I deleted it, so I need to delete this import too. Thanks for seeing it !

cgntnr: No worries !

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
Xelowak: True ! I changed it.

app/src/main/java/ch/epfl/sdp/blindly/match/UserListFilter.kt
Xelowak: Unfortunately, it wouldn't be possible since `getPotentialMatchesFromDatabase` queries a Firebase Firestore database (see in `MatchingAlgorithm.kt`), which is not possible in this class (because it uses a list of users instead).

cgntnr: I see, just wanted to check. Thanks for looking into it.

app/src/main/java/ch/epfl/sdp/blindly/match/UserListFilter.kt
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * @param userList the list to be filtered
+     * @return the same list without null users
+     */
+    fun clearNullUsers(userList: List<User?>): List<User> {
+        val nonNullList: MutableList<User> = ArrayList<User>().toMutableList()
+        for (user in userList) {
+            if (user != null) {
+                nonNullList += user
+            }
```
Lollipoke, line 83: I think you can use filterNotNull for that ^^
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/filter-not-null.html

app/src/main/java/ch/epfl/sdp/blindly/match/UserListFilter.kt
```diff
@@ -0,0 +1,105 @@
...
+        for (user in otherUsers) {
+            val otherUserAge = User.getUserAge(user)
+            if (otherUserAge != null) {
+                if (otherUserAge in minAge!!..maxAge!! && isLocatedInUserRadius(
+                        currentUser,
+                        user
+                    )
+                ) {
+                    filteredList += user
+                }
```
Lollipoke, line 35: I think you can use the filter method for that, to make it cleaner
For example : https://stackoverflow.com/questions/44518251/list-filter-within-the-outer-range-kotlin

app/src/main/java/ch/epfl/sdp/blindly/match/MatchingAlgorithm.kt
```diff
@@ -0,0 +1,73 @@
...
+            query.get().addOnSuccessListener { users ->
+                for (user in users) {
+                    matches += user.toUser()
+                }
+            }.addOnFailureListener { exception ->
+                Log.w(TAG, "Error getting users : ", exception)
+            }
+        }
+
+        val nonNullMatches = userListFilter.clearNullUsers(matches)
```
Lollipoke, line 69: Here you could use the filterNotNull method in one line, as I said in the other comment

app/src/androidTest/java/ch/epfl/sdp/blindly/match/UserListFilterTest.kt
```diff
@@ -0,0 +1,150 @@
...
+    @Test
+    fun filterLocationAndAgeRangeWorks() {
+        val alice = userBuilder.setUsername("Alice")
+            .setBirthday("29.04.1997")
+            .setAgeRange(listOf(20, 30))
+            .setLocation(locationEPFL)
+            .setRadius(50)
+            .build()
+
+        // No problem, Bob is perfect for Alice ! <3
```
Lollipoke, line 36: Yaaaay, gotta love the incoming love story <3

Lollipoke (commented):
I think it would be cleaner and easier to use the built-in filter methods for some function you wrote, especially if we aim for an efficient algorithm. Other than that, it looks clean, very well documented and well writen, great job !

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
```diff
@@ -81,7 +81,7 @@ class User private constructor(
          * @param location the location of the User
          */
         fun setLocation(location: List<Double>) = apply {
-            if(location.size == SIZE_2)
+            if (location.size == 2)
```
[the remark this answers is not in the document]
cgntnr: It was due to our previous exchange, he changed that to 2, because it was SIZE_2 which is not very meaningful either. I think we had a communication breakdown there what I meant was maybe to put something more explanatory like SIZE_OF_LOC_ATTRIBUTES etc. Because ```location.size == SIZE_2``` is not really better than ```location.size == 2```.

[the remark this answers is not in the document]
Lollipoke: yeah maybe the name could be more meaningful indeed

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MatchPageFragment.kt
```diff
@@ -173,6 +173,41 @@ class MatchPageFragment : Fragment(), CardStackListener {
...
+        } else {
+            createProfilesFromUsers(potentialUsers)
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    private fun createProfilesFromUsers(users: List<User>?): List<Profile> {
+        if (users == null) {
+            return listOf()
+        }
```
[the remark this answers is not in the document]
Xelowak: Yes it is, since the function takes a `List<User>?`, which can be null. Otherwise, I could use the "?" notation, but I think this null test handles this better.

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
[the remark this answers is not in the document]
Xelowak: At the end, I used two constants with differents names, namely `SIZE_OF_LOCATION_LIST` and `SIZE_OF_AGE_RANGE_LIST` which are both equal to 2.

app/src/main/java/ch/epfl/sdp/blindly/match/UserListFilter.kt
Xelowak: True, I didn't know something existed for that.

app/src/main/java/ch/epfl/sdp/blindly/match/UserListFilter.kt
Xelowak: Yep, I'll try it.

Lollipoke (approved):

PR #149, opened 2021-04-23
Age range

cmsigrist:
Add an age range field for the User, which can be used for the matching algorithm.


app/src/main/java/ch/epfl/sdp/blindly/profile_setup/ProfilePassions.kt
```diff
@@ -42,6 +44,7 @@ class ProfilePassions : AppCompatActivity() {
      *
      * @param view the current view
      */
+    @RequiresApi(Build.VERSION_CODES.O)
```
cgntnr, line 47: Maybe you can explain why these RequiresApi annotations are needed.

app/src/main/java/ch/epfl/sdp/blindly/user/UserHelper.kt
```diff
@@ -106,19 +110,28 @@ class UserHelper {
         return FirebaseAuth.getInstance().currentUser?.uid
     }
 
-    //TODO refactor this and move to UserRepository
     /**
      * Sets all the information passed by the userBuilder in Firebase Firestore
      *
      * @param userBuilder a builder from which a User can be build
```
cgntnr, line 116:
```suggestion
     * @param userBuilder a builder from which a User can be built
```

app/src/main/java/ch/epfl/sdp/blindly/user/UserHelper.kt
```diff
@@ -106,19 +110,28 @@ class UserHelper {
         return FirebaseAuth.getInstance().currentUser?.uid
     }
 
-    //TODO refactor this and move to UserRepository
```
cgntnr, line 109: Was this TODO no longer relevant, in terms of moving it to UserRepository?

cgntnr (approved):
Not much to say, I've added some comments that might be addressed. Accurate amount of changes for the PR this time :D Thanks!

app/src/main/java/ch/epfl/sdp/blindly/user/UserHelper.kt
cmsigrist: I thought a bit more about it and  I don't believe it would be very practical to move it to the UserRepository, since then it would need a @Injection userHelper, just to get the uid....
But then again it would make more sense to put it in the repo...

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
```diff
@@ -35,27 +36,29 @@ class User private constructor(
      * @property location the location of the User
      * @property birthday the birthday of the User
      * @property gender the gender of the User
-     * @property sexual_orientations an List<String> containing the
+     * @property sexualOrientations an List<String> containing the
```
Xelowak, line 39:
```suggestion
     * @property sexualOrientations a List<String> containing the
```

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
```diff
@@ -148,6 +151,17 @@ class User private constructor(
...
+         * Set the age range in the UserBuilder
+         *
+         * @param ageRange a list containing the age range :
+         *     ageRange[0] = minAge,
+         *     ageRange[1] = maxAge
+         */
+        fun setAgeRange(ageRange: List<Int>) = apply {
+            this.ageRange = ageRange
+        }
+
```
Xelowak, line 164: I think it may be useful to check for the size of the list before setting it to the user, so that we can avoid bugs with lists of more than 2 elements (and less than 2)

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
```diff
@@ -215,22 +233,26 @@ class User private constructor(
...
-                ).years
-
-                return age.toString()
+                return getAgeFromBirthday(birthday)
             }
             return null
         }
+
+        @RequiresApi(Build.VERSION_CODES.O)
+        fun getAgeFromBirthday(birthday: String): Int {
```
Xelowak, line 245: I might be wrong, but I think this function can be private.

Xelowak (approved):
That's nice ! I made some comments and suggestions that you might need to consider.

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
cmsigrist: It unfortunately cannot, I didn't know where to put this function but it's used in the setUser in the UserHelper (to compute the age range) and it didn't feel right to put this function in the UserHelper (it's already long enough) :')

cgntnr:
I think as a rule of thumb it would be nice to run ./gradlew connectedCheck before pushing the possibly final commit of a PR which would eventually show these errors. Since Cirrus is already taking too much time, it might save you some time.

cmsigrist:
> I think as a rule of thumb it would be nice to run ./gradlew connectedCheck before pushing the possibly final commit of a PR which would eventually show these errors. Since Cirrus is already taking too much time, it might save you some time.

True I thought about it when I saw the test failing the first time (though I was working on two branches at the same time which apparently wasn't such a good idea because some lines I thought I had changed on one were actually on the other branch x) )

cgntnr:
No problem! Thanks for the fix and additions :D

PR #150, opened 2021-04-23
Change test in profile page, they now work!

cmsigrist:
Changes the way the fragment was instanciated in the test.
And adds tests related to User and Fragment to increase the code coverage.


Xelowak (approved):
LGTM ! That's nice if you could find what was the problem 👍

Lollipoke (approved):
Nothing much to say, good job :)

PR #151, opened 2021-04-25
Move the init() and release() to Before and After

cmsigrist:
Has the title says, it moves the init() to @Before and the release() to a @After in each test


Xelowak (approved):
It's ok for me ! It could be nice if it stops making these useless "init was called twice in a row" after merging on main.

koko-ng (approved):
Well, I know I theoretically I should find something to say but after reviewing line by line, I couldn't find anything. So all good for me :)

PR #152, opened 2021-04-27
Chat

cgntnr:
- [x] Implement back-end for chat
- [x] Implement UI for chat
- [ ] Test


app/src/main/java/ch/epfl/sdp/blindly/chat/Message.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Empty constructor is needed for the Realtime Database.
+     */
+    constructor()
+
+    // timestamp is to achieve correct ordering of the messages and ensure fetching is in
+    // same ordering the next time users open the chat
+    // Note: I considered usage from different timezones however since matches are only
+    // allowed to be within a certain distance, there won't be any problem.
```
Lollipoke, line 13: What if both matches are close but on the limit of 2 timezones ?

app/src/main/res/layout/message_outgoing.xml
```diff
@@ -0,0 +1,25 @@
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/relativeLayout_outgoing"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content">
+
+    <RelativeLayout
```
Lollipoke, line 6: Cirrus says that the RelativeLayout or its parent might be useless, and that we could transfer the background attribute to the other view ?

app/src/androidTest/java/ch/epfl/sdp/blindly/chat/ChatAdapterTest.kt
```diff
@@ -0,0 +1,47 @@
...
+        val userId2 = "23094823049"
+        val m1 = Message("a message", userId1)
+        val m2 = Message("some message", userId2)
+        assertNotSame(m1.timestamp!!, m2.timestamp!!)
+        assertEquals("some message", m2.messageText)
+        assertEquals(userId2, m2.currentUserId)
+    }
+
+
+}
```
Lollipoke, line 47: Wouldn't it be better to use espresso methods instead of assertions ?

Lollipoke (approved):
Some trivial details, otherwise it seems good !

app/src/main/java/ch/epfl/sdp/blindly/chat/Message.kt
cgntnr: I'm willing to take the punishment if that happens :D Though while thinking about the problem I wonder what would happen if timezone is changed manually by hand. I will look into this.

app/src/main/res/layout/message_outgoing.xml
cgntnr: Yes, I tried to fix this one. Same warning was there for message_incoming and I could change that one but for outgoing without the extra layout it jumps back to the left side of the screen so I had to leave it.

app/src/main/java/ch/epfl/sdp/blindly/chat/ChatActivity.kt
```diff
@@ -0,0 +1,143 @@
...
+     * @param savedInstanceState
+     */
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_chat)
+        currentUserId = FirebaseAuth.getInstance().currentUser.uid
+
+        val bundle = this.intent.extras
+        matchId = if (bundle != null) {
+            bundle.getString("matchId").toString()
```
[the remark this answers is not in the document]
cgntnr: This was due to bundle being nullable right now. The else block is only there to test it by hand but ideally when chat is integrated with the matches we can provide bundle only with a valid match id and provide a non-null assertion. So then toString() won't be needed.
tl;dr It's only needed right now for the compilation.

app/src/androidTest/java/ch/epfl/sdp/blindly/chat/ChatAdapterTest.kt
cgntnr: Espresso is mostly for dealing with the views if I'm not mistaken but in these very basic tests I could only test some of the adapter. The whole instance of these classes are problematic as they reach out to Firebase so crashes the tests. We need to figure another way to test these.

app/src/main/java/ch/epfl/sdp/blindly/chat/Message.kt
cgntnr: Javadoc was unfortunately not very clear but upon further research I learnt that currentTimeMillis() is not timezone dependent as it gets the number of milliseconds passed since January 1st 1970 UTC. Noting that it's still local clock dependent so there could possibly be changes due to local clock set incorrectly but I think it's a bit too much nitpicking at this point.

Lollipoke: Yeah it is, I guess it's really not that important

.gitignore
```diff
@@ -1,5 +1,7 @@
 *.iml
+*.json
 .gradle
+/.idea
```
koko-ng, line 4: We shouldn't ignore all the idea folder, see [https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems)

app/src/androidTest/java/ch/epfl/sdp/blindly/chat/ChatAdapterTest.kt
koko-ng: Couldn't it be converted to a junit test (under test/ not androidTest/), especially as it doesn't use any espresso method as @Lollipoke mentioned.

app/src/main/java/ch/epfl/sdp/blindly/chat/ChatActivity.kt
```diff
@@ -0,0 +1,143 @@
...
+        } else {
+            //for manual testing
+            //matchId = "QDs8Hu6kWAb7SSdGQUbDllJXX3y2" //ct
+            "kq37EpNoqqPA3o6vjAFFuRUgDkE2" //cgngr
```
koko-ng, line 47: Why not just put the test id in the bundle why testing? I don't think having test code in production is a good idea.

koko-ng (changes requested):
Nicely done, thanks :)

app/src/androidTest/java/ch/epfl/sdp/blindly/chat/ChatAdapterTest.kt
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Test
+    fun itemTypeIsCorrect() {
+        val m1 = Message("mes1", "2938473dsgfd298")
+        val m2 = Message("mes2", "293s73dshfd298")
+        val list: ArrayList<Message> = arrayListOf<Message>()
+        list.addAll(listOf(m1, m2))
+        val abc: RecyclerView.Adapter<ChatAdapter.ViewHolder> = ChatAdapter(list)
+        assertEquals(1, abc.getItemViewType(1))
```
Xelowak, line 31: I think it might be nice to make the 1 a constant so that we understand that is it related to `REMOTE_USER_SENDING` (or make this constant in `ChatAdapter` accessible from other classes).

app/src/androidTest/java/ch/epfl/sdp/blindly/chat/ChatAdapterTest.kt
Xelowak: Using Espresso wouldn't be useful for these types of tests. 
However, I advise you to use Hamcrest as we did in other tests (using `assertThat `instead of `assertEquals `or `assertNotSame`)

app/src/androidTest/java/ch/epfl/sdp/blindly/chat/ChatAdapterTest.kt
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+
+    @Test
+    fun messageObjectIsCorrect() {
```
Xelowak, line 36: The name of the test is not really representative of what it does in my opinion. Maybe move the timestamp test of line 41 in another test function and consider only m2 in this test (as m1 is used only in the timestamp test).

Xelowak (approved):
Nice work ! I made some comments that you need to consider. Otherwise, I think it's ready to go ! :)

app/src/androidTest/java/ch/epfl/sdp/blindly/chat/ChatAdapterTest.kt
Lollipoke: Yeaaaah that's what I meant xD I mixed up the names

Lollipoke (approved):

app/src/main/java/ch/epfl/sdp/blindly/chat/ChatActivity.kt
cgntnr: You are right, I corrected this.

koko-ng (approved):
Perfect, Thanks :)

PR #153, opened 2021-04-27
Add location into firebase

r-ismer:


app/build.gradle
```diff
@@ -42,6 +42,9 @@ android {
         jvmTarget = '1.8'
     }
 
+    testOptions {
+        unitTests.returnDefaultValues = true
```
koko-ng, line 46: Is this really mandatory? Android docs seems to tell it should really be avoided, moreover I don't see any strange tests that could justify it

app/src/main/java/ch/epfl/sdp/blindly/location/AndroidLocationService.kt
```diff
@@ -1,21 +1,20 @@
...
+import android.location.Geocoder
 import android.location.Location
 import android.location.LocationManager
 
 private const val MIN_TIME_FOR_UPDATE = 1L
 private const val MIN_DISTANCE_FOR_UPDTAE = 1F
+private const val EPFL_LAT = 46.5
+private const val EPFL_LONG = 6.5
 
-/**
```
koko-ng, line 10: Where is it gone? :(

app/src/main/java/ch/epfl/sdp/blindly/profile_setup/ProfilePassions.kt
```diff
@@ -74,9 +73,6 @@ class ProfilePassions : AppCompatActivity() {
         }
     }
 
-    /**
```
koko-ng, line 77: Please, stay comments, we will really miss you.

app/src/main/java/ch/epfl/sdp/blindly/settings/SettingsLocation.kt
```diff
@@ -20,7 +20,7 @@ class SettingsLocation : AppCompatActivity(), OnMapReadyCallback {
     private lateinit var mapView: MapView
     private var location: Location? = null
     private lateinit var locSer: LocationService
-    private val LAUSANNE = LatLng(46.5, 6.6)
+    private val LAUSANNE_LATLNG = LatLng(46.5, 6.6)
```
koko-ng, line 23: I guess we are using Lausanne and EPFL as default values? Maybe we should merge them somehow, seems a little bit odd to me to have to different defaults.

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
```diff
@@ -17,14 +17,14 @@ import java.time.Period
...
-    val location: String?,
+    val location: List<Double>?,
     val birthday: String?,
     val gender: String?,
     val sexualOrientations: List<String>,
     val showMe: String?,
     val passions: List<String>,
     val radius: Int?,
-    val matches: List<User>,
+    val matches: List<String>,
```
koko-ng, line 27: It is not very clear, why is the changed? Is `matches` now a list of IDs? If it is the case, I think renaming to something like matchesIds could be clearer.

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
```diff
@@ -200,14 +199,14 @@ class User private constructor(
         fun DocumentSnapshot.toUser(): User? {
             try {
                 val username = getString("username")!!
-                val location = getString("location")!!
+                val location = get("location") as List<Double>
```
koko-ng, line 202: I think it is a bit risky to do a unsafe cast here, imagine a user put an invalid list of Strings in the db, could they make every app that encounters them while swiping crash? Maybe a safe cast is the way to go: [https://kotlinlang.org/docs/typecasts.html#safe-nullable-cast-operator](https://kotlinlang.org/docs/typecasts.html#safe-nullable-cast-operator)

koko-ng (changes requested):
Overall it seems nice, despite some small points. I also thinks that your linter went a bit crazy and deleted a lot of comment, let me know if it was intentional.

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
```diff
@@ -76,7 +76,7 @@ class User private constructor(
          *
          * @param location the location of the User
          */
-        fun setLocation(location: String) = apply {
+        fun setLocation(location: List<Double>) = apply {
```
cmsigrist, line 79: Maybe you should add checks on the input length (that it only contains two doubles)

app/src/main/java/ch/epfl/sdp/blindly/settings/SettingsLocation.kt
```diff
@@ -46,7 +46,7 @@ class SettingsLocation : AppCompatActivity(), OnMapReadyCallback {
         val loc: LatLng = if (location != null) {
             LatLng(location!!.latitude, location!!.longitude)
         } else {
-            LAUSANNE
+            LAUSANNE_LATLNG
```
cmsigrist, line 49: I think as Tibor said that we don't need the LAUSANNE_LATLING, even more so if it's only used as a default value, maybe it would be better to have an empty list then?

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
cmsigrist: I don't know, to me it makes more sense to check the inputs, so that no invalid inputs are put on the database in the first place

cmsigrist (commented):
Great! It's indeed a lot better to have the latitude and longitude rather than a String
I'll approve once you've #BringBackTheComments:'(

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
cgntnr: It was actually discussed in the meetings, this is preferred not to unnecessarily store whole user for each match where as our matching algorithm only needs UIDs which are stored as Strings.

app/build.gradle
r-ismer: I had to because the unit tests won't pass without it

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
koko-ng: Of couse but a malicious user could still send queries from outside the app to put invalid data in our db, especially as we aren't doing any kind of server-side validation of the data.

app/build.gradle
r-ismer: i checked on android dev https://developer.android.com/training/testing/unit-testing/local-unit-tests.html

koko-ng (approved):

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
```diff
@@ -9,24 +9,25 @@ import kotlinx.serialization.Serializable
 import java.time.LocalDate
 import java.time.Period
 
+const val SIZE_2 = 2
```
cgntnr, line 12: I'm not really sure of this naming, what is the point of having the constant :D Maybe something like SIZE_LAT_AND_LNG ? I leave it up to you.

cmsigrist (approved):
Nice work! :)

cgntnr (approved):
A small remark on a constant otherwise the other comments should be enough. Thanks!

PR #154, opened 2021-04-27
Firebase emulator

Xelowak:
Initialized a Firebase CLI project for our project, so that we can emulate the database for test.


Xelowak:
After a lot of web browsing, I couldn't find something that was persistent. We're still searching for a solution, but this branch doesn't seem to resolve the issue for now.

PR #155, opened 2021-04-28
Shorten audio duration in tests

koko-ng:
Allow the recorded audio duration to be shortened during tests.


Xelowak (approved):
It's ok for me

Lollipoke (approved):

PR #159, opened 2021-04-30
Refactor match screen

Lollipoke:
Title says it all :p
We might want to find a way to disable swiping through the different fragments as it is hard to swipe matches tho


app/src/main/java/ch/epfl/sdp/blindly/main_screen/MatchPageFragment.kt
```diff
@@ -59,18 +72,81 @@ class MatchPageFragment : Fragment() {
...
+    }
+
+    /**
+     * Initialize the manager, the adapter and the cardStackView
+     *
+     */
+    private fun setupCardStackView(view: View) {
+        cardStackView = view.findViewById(R.id.card_stack_view)!!
+        manager = CardStackLayoutManager(context, this)
+        manager.setStackFrom(StackFrom.None)
```
cmsigrist, line 129: Could all these calls to set() not be chained?

r-ismer (approved):
It's very clear to me because I wrote the match activity. The integration in the fragment is well done !! Good job

cmsigrist (approved):
This refactor looks all good to me! Great job!

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MatchPageFragment.kt
Lollipoke: Unfortunately no, because those methods don't return a CardStackLayoutManager so you can't link them :(

PR #169, opened 2021-04-30
Fix permission test

koko-ng:
I deleted something in a test that I think wasn't useful any more, but once merged in main it is still needed for an unclear reason. It appears to also be necessary for another test.


Lollipoke (approved):
Yaaaay I hope it will fix our problems

cgntnr (approved):

PR #170, opened 2021-04-30
Fix cirrus

cgntnr:
I tried another thing to fix cirrus problems not sure if it will work on Cirrus. We can merge if it's okay the first time it runs on Cirrus.


cgntnr:
Nope, apparently still not working.

PR #173, opened 2021-05-01
Fix settings tests intents

koko-ng:
Fixes some tests we missed while refactoring init/release intents.

Merge after #174


cmsigrist (approved):
Great! That should make the run checks on cirrus a lot smoother!

cgntnr:
I see that many of the changes here were already in #174. I'll let you merge main into this to avoid any wrong doing and then review if that's okay. @koko-ng

cgntnr (approved):
Good to go! Many thanks again for the fix.

PR #174, opened 2021-05-01
Fix permission rule race cond

koko-ng:
Fixes the terrible race condition that made all our pipelines sad. `GrantPermissionRule` takes a bit of time to apply unfortunately.


Lollipoke (approved):

cgntnr (approved):
Let's hope it works this time consistently, thank you very much for the hard work!

PR #176, opened 2021-05-01
Build and test only for the debug variant

koko-ng:
Changes: 
 - Compile only the debug variant for the tests, the checks on cirrus are around one minute shorter from my tests
  - Upload the tests reports as artifacts. If a test fails on cirrus we therefore have the full stacktrace available for easier debug. I also think that cirrus indicates which tests failed on the associated PR however we will see if this really happens next time a pipeline fails :)

**EDIT:** The pipeline failed and cirrus doesn't report which test failed, but having the results in cirrus is a plus anyway IMO


Xelowak (approved):

cgntnr:
I know this has been been discussed between you @koko-ng  and the TAs but could you explain why you replaced connectedCheck. I haven't seen this change in any of the groups, the first build failed and there doesn't seem to be an advantage in terms of duration; 13 mins as it was mostly around that. Maybe we shouldn't mess around with Cirrus after you put so much work in it to fix. I definitely agree that we should add --stacktrace to connectedCheck though.

koko-ng:
Well basically what we do now is run two gradle tasks, check and connectedCheck. We do this in order to run all our tests and generate the coverage report. 
Our app has two build variant, debug and release, which depend on a plethora of subtasks, some which are very similar but still executed twice. You can see it from the logs with, for instance, mergeDebugResources and mergeReleaseResources. The problem with check and connectedCheck is that the build variant is not specified so some task that get executed are not useful to us, for instance generating the release build. Running only jacocoTestReport runs _only_ the necessary tasks for the coverage, which are the tests. You can see it in settings.gradle:145 : `task jacocoTestReport(type: JacocoReport, dependsOn: ['testDebugUnitTest', 'createDebugCoverageReport'])`.
Effectively it doesn't reduce the execution time by a large factor but I think it is always worth taking it.

By the way it wasn't in fact a problem with Cirrus but more getting around a race condition in espresso.
By the way 2, the generated stack traces are generated even without --stacktrace, which is only for generating gradle stack trace, which we don't need because gradle is definitely a better piece of software than espresso. 😬

cgntnr:
Fair enough!

cgntnr (approved):

PR #177, opened 2021-05-02
Fix going back after setting profile

cgntnr:
Having finished the profile setup (when in MainScreen) back button press prompts a message, asks the user to exit the app or not. This was needed not only after registering but also when a registered users signs into app. Because otherwise back button would send you back to splash screen and gets stuck there.

TODO : I realized that same problem is there if we press back button when signing up or signing in however there's not an activity that I can override ```backButtonPressed()``` because API handles the intent with ```getSignInIntent()```. Any idea is welcome on that front, otherwise this PR fixes the bug once profile setting is finished and once a user signs in.

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/blindly/main_screen/MainScreen.kt
+++ b/app/src/main/java/ch/epfl/sdp/blindly/main_screen/MainScreen.kt
@@ -1,5 +1,6 @@
 package ch.epfl.sdp.blindly.main_screen
 
+import android.app.AlertDialog
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
 import androidx.viewpager2.widget.ViewPager2
@@ -9,6 +10,7 @@ import ch.epfl.sdp.blindly.R
 import com.google.android.material.tabs.TabLayout
 import com.google.android.material.tabs.TabLayoutMediator
 import dagger.hilt.android.AndroidEntryPoint
+
 
 /**
  * This activity holds the three fragments (Match, Message and Profile page)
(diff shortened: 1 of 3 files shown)


app/src/main/java/ch/epfl/sdp/blindly/main_screen/MainScreen.kt
```diff
@@ -45,4 +47,21 @@ class MainScreen : AppCompatActivity() {
             }
         })
     }
+
+    override fun onBackPressed() {
+        val builder: AlertDialog.Builder = AlertDialog.Builder(this)
+        builder.setTitle("Exit")
+        builder.setMessage("Are You Sure?")
```
Lollipoke, line 54: Does the message say anything else than "Are you Sure?" ? I think it could be better to inform the user that doing so will close the app. Otherwise it might be taken as a "Are you sure you want to go back to the sign up part?"

Lollipoke (approved):
Short but efficient, good job :)

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MainScreen.kt
cgntnr: Check the line 53 :D But if you want, we can further specify further as "Exit the app." instead of Exit.

Lollipoke: Haha yeah, mb! But I think it would be clearer to add the app :D

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MainScreen.kt
```diff
@@ -45,4 +47,21 @@ class MainScreen : AppCompatActivity() {
...
+    override fun onBackPressed() {
+        val builder: AlertDialog.Builder = AlertDialog.Builder(this)
+        builder.setTitle("Exit the app.")
+        builder.setMessage("Are You Sure?")
+        builder.setPositiveButton("Yes") { dialog, _ ->
+            dialog.dismiss()
+            this.finishAffinity()
+        }
+        builder.setNegativeButton(
+            "No"
```
Xelowak, line 60: I think it may be nice to have those strings stored in constants. It could be better to use @string resource, but I don't think it can be used there and that's why I suggest using constants.

Xelowak (approved):
Nice ! This way we won't have bugs with users registering multiplte times. I made a comment regarding string constants. Feel free to tell me if you think it's useless to do so.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MainScreen.kt
cgntnr: Sure, always a good practice to use constants, I will address that.

PR #180, opened 2021-05-05
Fetch potential matches from database

Lollipoke:
I used Kilian's matching algorithm to fetch potential matches directly from the database instead of having a hardcoded list of profile.
I also fixed the type of ageRange in User.kt

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/blindly/match/MatchingAlgorithm.kt
+++ b/app/src/main/java/ch/epfl/sdp/blindly/match/MatchingAlgorithm.kt
@@ -1,6 +1,7 @@
 package ch.epfl.sdp.blindly.match
 
 import android.os.Build
+import android.util.Log
 import androidx.annotation.RequiresApi
 import ch.epfl.sdp.blindly.user.User
 import ch.epfl.sdp.blindly.user.User.Companion.toUser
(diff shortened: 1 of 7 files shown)


cgntnr (approved):
I couldn't notice anything to add, good job. Though we don't have the means to test via Emulator etc. I'm guessing you checked their functioning either by hand or with local tests.

app/src/main/java/ch/epfl/sdp/blindly/match/MatchingAlgorithm.kt
```diff
@@ -49,23 +47,19 @@ class MatchingAlgorithm {
...
-            query.get().addOnSuccessListener { users ->
-                for (user in users) {
+            //Wait on the query to be done before continuing
+            val users = query.get().await()
+            for (user in users) {
+                if (user.id != userid) {
                     matches += user.toUser()
                 }
-            }.addOnFailureListener { exception ->
-                Log.w(TAG, "Error getting users : ", exception)
```
Xelowak, line 65: What happens now if the query fails ?

Xelowak (approved):
Everything looks fine to me.

app/src/main/java/ch/epfl/sdp/blindly/match/MatchingAlgorithm.kt
Lollipoke: You're right, I added a try-catch :)

PR #182, opened 2021-05-05
Live map and ChatActivity tests

koko-ng:
Add the live map to see where another user is, while sending them your location.

I also too the opportunity to make a common class for the live database, I could therefore test ChatActivity.


cgntnr:
Hello,  just wanted to let you know the app crashes when you click on User Map button on map fragment. Was this expected for any reason? By the way very good job on the refactoring and tests, greatly appreciated. One other thing is that many unchanged lines seems to be deleted and re-added, makes it hard to follow which lines you added or which lines were already there. Do you have any idea why?

koko-ng:
> Hello, just wanted to let you know the app crashes when you click on User Map button on map fragment. Was this expected for any reason? 
Seems strange, it works on my side and there is a test checking that opening the user map from the main screen works. Is it still the case?
> By the way very good job on the refactoring and tests, greatly appreciated. 
Thanks, no problem :)
>One other thing is that many unchanged lines seems to be deleted and re-added, makes it hard to follow which lines you added or which lines were already there. Do you have any idea why?
Yes I messed up with git autocrlf, but I fixed it in the latest commits I think. On my side everything seems ok at [https://github.com/BlindlyTeam/Blindly/pull/182/files](https://github.com/BlindlyTeam/Blindly/pull/182/files).
Let me know if it isn't the case on your side.

cgntnr:
> Seems strange, it works on my side and there is a test checking that opening the user map from the main screen works. Is it still the case?
I've created a new account and afterwards, it was fine.

app/src/androidTest/java/ch/epfl/sdp/blindly/UserMapTest.kt
```diff
@@ -4,49 +4,137 @@ import android.content.Intent
...
+            }
+        })
+        assertEquals(completed, future.get())
+    }
     @Test
-    fun openAndDisplayUserWithoutPoints() {
-        val intent = Intent(ApplicationProvider.getApplicationContext(), UserMapActivity::class.java)
+    fun receiveInvalidLocation() {
+        // Constant to check future constant completion
+        val completed = "Location received"
```
cmsigrist, line 107: Maybe put this const in the class rather than in each test

app/src/androidTest/java/ch/epfl/sdp/blindly/UserMapTest.kt
```diff
@@ -4,49 +4,137 @@ import android.content.Intent
...
+        database.getLocationLiveDatabase(user.getUserId()!!, OTHER_USER_ID).addListener(object :
+            DatatbaseHelper.BlindlyLiveDatabase.EventListener<BlindlyLatLng>() {
+            override fun onMessageReceived(message: Message<BlindlyLatLng>) {
+                future.complete(completed)
+            }
+            override fun onMessageUpdated(message: Message<BlindlyLatLng>) {
+                future.complete(completed)
+            }
+        })
+        assertEquals(completed, future.get())
```
Xelowak, line 102: It might be better to use Hamcrest's `assertThat` since we use it in other tests (and it is clearer). Be careful if you commit my suggestion, you might need some imports.
```suggestion
        assertThat(future.get(), equalTo(completed))
```

app/src/androidTest/java/ch/epfl/sdp/blindly/UserMapTest.kt
Xelowak: I agree !

app/src/androidTest/java/ch/epfl/sdp/blindly/UserMapTest.kt
```diff
@@ -4,49 +4,137 @@ import android.content.Intent
...
+                override fun onMessageUpdated(message: Message<BlindlyLatLng>) {
+                    future.complete(completed)
+                }
+            })
+            // Invalid data
+            liveDb.updateLocation(BlindlyLatLng())
+        }
+        // We wait for the future to complete without other assert, because
+        // we want to check that the activity doesn't crash
+        assertEquals(completed, future.get())
```
Xelowak, line 137: Same comment about assertEquals.

app/src/androidTest/java/ch/epfl/sdp/blindly/chat/LiveChatTest.kt
```diff
@@ -0,0 +1,122 @@
...
+        return Predicate<View?> { view -> matcher.matches(view) }
+    }
+
+    private fun sendMessage() {
+        onView(withId(R.id.newMessageText))
+            .perform(clearText(), typeText(TEST_MESSAGE))
+
+        closeSoftKeyboard()
+        val buttonUpdate = onView(withId(R.id.sendButton))
+        buttonUpdate.check(matches(withEffectiveVisibility(ViewMatchers.Visibility.VISIBLE)));
```
Xelowak, line 94: It may be a good idea to import directly `ViewMatchers.Visibility.VISIBLE`, so that is is less redundant to call.

app/src/androidTest/java/ch/epfl/sdp/blindly/chat/LiveChatTest.kt
```diff
@@ -0,0 +1,122 @@
...
+    fun sendReceiveWorks() {
+        sendMessage()
+        onView(withId(R.id.recyclerView)).check(matches(withViewCount(withText(TEST_MESSAGE), 2)))
+    }
+
+    @Test
+    fun sendReceiveMultipleMessagesWorks() {
+        for (i in 0..1) {
+            sendMessage()
+        }
```
Xelowak, line 107: Is a for loop really needed for 2 occurences ? I think that writing `sendMessage()` 2 times is better for this case.

app/src/main/java/ch/epfl/sdp/blindly/UserMapActivity.kt
```diff
@@ -35,8 +54,56 @@ class UserMapActivity: AppCompatActivity(), OnMapReadyCallback, ActivityCompat.O
         val mapFragment = supportFragmentManager
                 .findFragmentById(R.id.map) as SupportMapFragment?
         mapFragment!!.getMapAsync(this)
+
+        // Cancel loading if we can't get the user id
+        val currentUserId = userHelper.getUserId() ?: return
+        val matchId = intent.extras?.getString(MATCH_ID) ?: "default_user"
```
cmsigrist, line 60: Sorry, due to my lack of thorough understanding of how this map works, I'll have to leave comments on trivial things...
I guess you could use a const String for the default_user, but feel free to ignore this comment........ x)

app/src/main/java/ch/epfl/sdp/blindly/di/DatabaseHelperModule.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+@Module
+@InstallIn(SingletonComponent::class)
+object DatabaseHelperModule {
+
+    /**
+     * Return a UserHelper to be injected
+     *
+     * @return UserHelper
+     */
```
Xelowak, line 21: I think you forgot to change the comments after a copy/paste.

app/src/main/java/ch/epfl/sdp/blindly/helpers/BlindlyLatLng.kt
```diff
@@ -0,0 +1,44 @@
...
+        this.latitude = latLng?.latitude
+        this.longitude = latLng?.longitude
+    }
+
+    fun toLatLng(): LatLng? {
+        latitude ?: return null
+        longitude ?: return null
+        return LatLng(latitude!!, longitude!!)
+    }
+}
```
Xelowak, line 44: To be honest I don't really understand the purpose of this class. Isn't it exactly the same as a LatLng object that already exists in Kotlin ?

app/src/main/java/ch/epfl/sdp/blindly/chat/ChatActivity.kt
```diff
@@ -66,23 +69,16 @@ class ChatActivity : AppCompatActivity() {
      * @param view the current view
      */
     fun sendButtonActivate(view: View) {
-        findViewById<ImageView>(R.id.sendButton).setOnClickListener {
-            if (findViewById<EditText>(R.id.newMessageText).text.toString().isNotEmpty()) {
-                sendMessage()
-            }
+        if (findViewById<EditText>(R.id.newMessageText).text.toString().isNotEmpty()) {
```
cmsigrist, line 72: Since you removed the .setOnClickListener, is it directly mapped to the button in the .xml?

app/src/main/java/ch/epfl/sdp/blindly/di/DatabaseHelperModule.kt
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * UserHelper Module fore dependency injection
+ */
+@Module
+@InstallIn(SingletonComponent::class)
+object DatabaseHelperModule {
+
+    /**
+     * Return a UserHelper to be injected
```
cmsigrist, line 18: Typo from copy/paste ;p

app/src/main/java/ch/epfl/sdp/blindly/helpers/Message.kt
```diff
@@ -10,17 +15,16 @@ class Message {
     // timestamp is to achieve correct ordering of the messages and ensure fetching is in
     // same ordering the next time users open the chat
     var timestamp: Long? = null
-    var messageText: String? = null
+    var messageText: T? = null
```
Xelowak, line 18: Is there a reason why the message contains a generic type now ? I guess it allows us to do more things with it, but since the message will only contain a String, is it really important to do so ?

app/src/main/java/ch/epfl/sdp/blindly/UserMapActivity.kt
```diff
@@ -35,8 +54,56 @@ class UserMapActivity: AppCompatActivity(), OnMapReadyCallback, ActivityCompat.O
...
+    }
+
+    private fun updateMarker(message: Message<BlindlyLatLng>) {
+        if (!::map.isInitialized) return
+        if (message.currentUserId == userHelper.getUserId()) return
+        val pos = message.messageText ?: return
+        val latLng = pos.toLatLng() ?: return
+        tryToAddMarker()?.position = latLng
+        // Always keep match position on the center of the map
+        map.moveCamera(CameraUpdateFactory.newLatLng(latLng))
```
Xelowak, line 89: Isn't it a better idea to keep ourselves at the center of the map ?

app/src/main/java/ch/epfl/sdp/blindly/helpers/DatatbaseHelper.kt
```diff
@@ -0,0 +1,206 @@
...
+     * @param userId the logged-in user id
+     * @param otherUserId the other user's id to send message to
+     * @return The live database
+     */
+    fun getChatLiveDatabase(
+        userId: String,
+        otherUserId: String
+    ): ChatLiveDatabase {
+        return ChatLiveDatabase(
+            databaseInstance.getReference("messages").child(getConversationId(userId, otherUserId)),
```
cmsigrist, line 45: Again maybe use const for "message" and "location" (beneath)

app/src/main/java/ch/epfl/sdp/blindly/helpers/DatatbaseHelper.kt
```diff
@@ -0,0 +1,206 @@
...
+     * get an instance
+     *
+     * @param dr the firebase database reference
+     * @param userId the logged-in user id
+     */
+    abstract class BlindlyLiveDatabase<T> internal constructor(protected val dr: DatabaseReference,
+                                                               protected val userId: String
+    ) {
+        // Init at one as most of the time we don't register more than one listener per db
+        private val eventListeners: MutableList<EventListener<T>> = ArrayList(1)
```
cmsigrist, line 127: Magic number... (sorry this is really nitpicking)

app/src/main/java/ch/epfl/sdp/blindly/helpers/Message.kt
cmsigrist, line 18: Shouldn't the type to be sent inside the message always be a String?

app/src/main/res/layout/activity_chat.xml
```diff
@@ -42,15 +42,16 @@
 
         <EditText
             android:id="@+id/newMessageText"
-            android:layout_width="339dp"
+            android:layout_width="fill_parent"
```
Xelowak, line 45: Cirrus says that is it better to use layout_width of 0dp for performance reasons.
```suggestion
            android:layout_width="0dp"
```

app/src/test/java/ch/epfl/sdp/blindly/MessageTests.kt
```diff
@@ -0,0 +1,51 @@
...
+    @Test
+    fun emptyConstructor() {
+        val msg = Message<BlindlyLatLng>()
+        val timestamp = System.currentTimeMillis()
+        msg.messageText = LOCATION_MESSAGE
+        msg.currentUserId = USER_ID
+        msg.timestamp = timestamp
+        assertEquals(LOCATION_MESSAGE, msg.messageText)
+        assertEquals(USER_ID, msg.currentUserId)
+        assertEquals(timestamp, msg.timestamp)
```
Xelowak, line 49: Again, it may be better to use assertThat for consistency and clarity.

cmsigrist (approved):
These are some awesome changes, good job!

Xelowak (approved):
Good work ! This quite a big PR and I think that it could have been nice to split it into 2 PRs : one for the chat changes and one for the live map. 
I made come comments that you need to consider, but I think the code is soon ready to be merged and that's why I approve it.
I really appreciate the fact that you took care of commenting your code ! :)

app/src/main/java/ch/epfl/sdp/blindly/UserMapActivity.kt
cgntnr: In fact this seems to be same code I used for Chat, what it does is that when it creates an entry in the Realtime Database it names the other user as "default_user as a title, so instead of an error it just creates a trivial entry, I can't say it's the best way but it seems safer. For location, I don't know if it might cause problems if you try to fetch from "default_user" because it won't have any location. I'll let @koko-ng reply that. In terms of making it constant, certainly possible :D

app/src/main/java/ch/epfl/sdp/blindly/helpers/Message.kt
cgntnr: This caught my eyes as well, I think @koko-ng decided to use same object to store locations as well. When I checked the Realtime Database I saw that locations were sent in a messageText and then it had latitude and longitude as child attributes. This might be better in terms of not replicating code and easier to test however it creates a little bit of confusion. Make we can adjust the naming a bit ?

cgntnr: See my comment on @cmsigrist 's remark https://github.com/BlindlyTeam/Blindly/pull/182#discussion_r627555824

app/src/main/java/ch/epfl/sdp/blindly/helpers/BlindlyLatLng.kt
koko-ng: It is very similar indeed, however Firebase had troubles receiving a LatLng objects so I wrapped it. Moreover I had a lot of code that needed Lat/Lng data in different formats (Location, LatLng, Doubles) so having a single class do everything avoid some boilerplate code. 
Furthermore the LatLng comes from the Google API, I think it is arguably better to wrap external classes that we depend on, especially given the stability of the APIs on android.

app/src/main/java/ch/epfl/sdp/blindly/helpers/Message.kt
koko-ng: No it won't the LiveLocation database sends a BlindlyLatLng.

app/src/androidTest/java/ch/epfl/sdp/blindly/chat/LiveChatTest.kt
koko-ng: Mmh why? Readability? It is a rather personal opinion haha.

app/src/androidTest/java/ch/epfl/sdp/blindly/UserMapTest.kt
koko-ng: Thanks, I didn't know about this. It is nice! First syntax is really ugly haha.

app/src/main/java/ch/epfl/sdp/blindly/chat/ChatActivity.kt
koko-ng: It always has been, however how the listener was implemented deactivated the button before the first click, which was causing troubles with the test. I don't know why it was implemented like that in the first place, maybe @cgntnr you have more insight on this?

cgntnr:
I had another question actually but maybe it's a task for another PR. At this moment it seems that as soon as there's a match both parties can see where other one is (as soon as matchedId is passed to userMap. This would include their job, school and even house location. So just a TODO for us that user map should only be available when both parties would like to find each other.

app/src/main/java/ch/epfl/sdp/blindly/helpers/DatatbaseHelper.kt
koko-ng: Well it is just one and doesn't change anything to the functionality. It is just a (very little) performance tuning

app/src/main/java/ch/epfl/sdp/blindly/helpers/Message.kt
koko-ng: No, the LiveLocation sends BlindlyLatLng

app/src/androidTest/java/ch/epfl/sdp/blindly/chat/LiveChatTest.kt
Xelowak: True. Feel free to do what you prefer.

app/src/main/java/ch/epfl/sdp/blindly/chat/ChatActivity.kt
cgntnr: I think the first click being deactivated is about focusable attribute. In terms of sendButton yes, the xml file already had the onClick attribute which called this function and if the input(newMessageText) is not empty calls the sendMessage to send the text to database and set the view. I didn't see any disadvantage really, less boilerplate code imo but I'm open to discussion/change if you feel otherwise.

app/src/main/java/ch/epfl/sdp/blindly/helpers/BlindlyLatLng.kt
Xelowak: Ok, thanks for the explanation !

Xelowak:
> I had another question actually but maybe it's a task for another PR. At this moment it seems that as soon as there's a match both parties can see where other one is (as soon as matchedId is passed to userMap. This would include their job, school and even house location. So just a TODO for us that user map should only be available when both parties would like to find each other.

Yeah you're right. We might need to address that later on.

app/src/main/java/ch/epfl/sdp/blindly/UserMapActivity.kt
Xelowak: Yep, that's a good idea ! This way the app can show us where to go at the beginning, and then we're free to move the map.

koko-ng:
> I had another question actually but maybe it's a task for another PR. At this moment it seems that as soon as there's a match both parties can see where other one is (as soon as matchedId is passed to userMap. This would include their job, school and even house location. So just a TODO for us that user map should only be available when both parties would like to find each other.

No, the location is updated only when the user has the map looking for the other user open, i.e. when the activity is open with the other userId given as argument.
However it could be interesting to add a feature to show that the location has not been updated since n minutes

app/src/main/java/ch/epfl/sdp/blindly/UserMapActivity.kt
koko-ng: Hahaha no problem, it seems legit.

app/src/main/java/ch/epfl/sdp/blindly/chat/ChatActivity.kt
koko-ng: Yes, but setting the correct click listener after the first click requires to click to send a message on the first time.
Like:
-  click 1:
    - set the click listener to (not empty -> send message)
-  click 2:
    - calls (not empty -> send message)
      - if not empty, send message

koko-ng:
> Good work ! This quite a big PR and I think that it could have been nice to split it into 2 PRs : one for the chat changes and one for the live map.
> I made come comments that you need to consider, but I think the code is soon ready to be merged and that's why I approve it.
> I really appreciate the fact that you took care of commenting your code ! :)

Yes I thought it would have been to split it but it was too late :(
Anyway thanks for the compliments :)

app/src/main/java/ch/epfl/sdp/blindly/chat/ChatActivity.kt
cgntnr: I think this [answer](https://stackoverflow.com/questions/46289043/android-onclicklistener-not-working-for-first-click-in-recyclerview) could solve the issue.

app/src/main/java/ch/epfl/sdp/blindly/UserMapActivity.kt
koko-ng: Yes, the pins are different.

app/src/main/java/ch/epfl/sdp/blindly/chat/ChatActivity.kt
koko-ng: I think I'm not really understanding what your are trying to do. Why set the actual click listener only after the second click? I don't think your code match your comment either, it is not 
>  checks the input EditText and if it's empty, calls the sendMessage.

but rather
>  attach a new click listener so that next time the button is click, it checks the input EditText and if it's empty, calls the sendMessage.

Anyway, do you think that my change breaks something you wanted to achieve, or can I merge ?
Sorry if I'm a bit confused about this part.

cgntnr: It calls only if it's not empty! To  my understanding there wasn't a second onListener but maybe I skipped something. Well no worries I think it got a bit confusing, but I tested out your code and seemed working fine so no worries go ahead and merge.

PR #179, opened 2021-05-05
Settings update

cmsigrist:
Links the settings of the user to the database, the user can now update his information.

PR #184, opened 2021-05-06
Add recording to matching

Lollipoke:
You can now listen to audio files on the swipe cards :D
There's still some work to do on the recordings like stop it when you swipe, pause it, disable multiple hearing and so on, but it's functional that I'll do in another PR
Tested it by hand tho, since I don't know how to mock the firebase storage
I added the gender and distance on the cards, and modified the middle button to play the audio instead of rewinding


app/src/main/java/ch/epfl/sdp/blindly/main_screen/MatchPageFragment.kt
```diff
@@ -50,6 +51,9 @@ class MatchPageFragment : Fragment(), CardStackListener {
     @Inject
     lateinit var userRepository: UserRepository
 
+    @Inject
+    lateinit var storage: FirebaseStorage
+
     companion object {
         private const val ARG_COUNT = "matchArgs"
         private var counter: Int? = null
```
cgntnr, line 59: I know this is from a previous PR but I wonder why a counter was initialized as null.

app/src/main/java/ch/epfl/sdp/blindly/match/Profile.kt
```diff
@@ -32,7 +36,13 @@ data class Profile(
     /**
      * A constructor for a profile from a parcel
      */
-    constructor(source: Parcel) : this(source.readString()!!, source.readInt())
+    constructor(source: Parcel) : this(
+        source.readString()!!,
+        source.readInt(),
```
cgntnr, line 41: Sorry if this is basic but at this point I'm a bit distant to this part of the code. Why the second element of the constructor doesn't have a non-null assertion operator? In any way, !!s might cause us some problems afterwards but maybe that was the only way.

app/src/main/res/layout/item_profile.xml
```diff
@@ -7,41 +7,75 @@
...
+    app:cardUseCompatPadding="true">
 
     <ImageView
         android:id="@+id/item_image"
-        android:contentDescription="SoundWaves"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        android:scaleType="centerCrop"/>
+        android:background="@color/blindly_blue"
+        android:contentDescription="SoundWaves"
```
cgntnr, line 19: We can add this to string resources instead of hardcoded string.

app/src/main/res/layout/item_profile.xml
```diff
@@ -7,41 +7,75 @@
...
-            android:textSize="20sp"/>
+            android:layout_marginBottom="15dp"
+            android:textSize="20sp"
+            android:textStyle="bold" />
+
+        <TextView
+            android:id="@+id/item_description"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:justificationMode="inter_word"
```
cgntnr, line 47: For some functions we had RequiresAPI annotation but not sure how having a API level 26 and higher might have an effect in a xml file. But I guess we can leave it if it's not replicable otherwise.

cgntnr (approved):
I added couple of comments you might want to look at, I believe we should at least merge #182 and #179 before this one.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MatchPageFragment.kt
```diff
@@ -224,7 +234,13 @@ class MatchPageFragment : Fragment(), CardStackListener {
...
         for (user in users) {
             profiles.add(
-                Profile(user.username!!, User.getUserAge(user)!!)
+                Profile(
+                    user.username!!,
+                    User.getUserAge(user)!!,
+                    user.description!!,
+                    user.passions!!.joinToString(", \n"),
+                    user.recordingPath!!
+                )
```
Xelowak, line 243: I was wondering something when I worked on the Matching Algorithm. Wouldn't it be better to have a function called `createProfileFromUser(user: User)` in `Profile.kt` ? It would be possible to test it and it seems more intuitive to me. However, I know it works like that.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MatchPageFragment.kt
```diff
@@ -270,4 +286,21 @@ class MatchPageFragment : Fragment(), CardStackListener {
...
+    /**
+     * Prepares the button to play an audio
+     *
+     */
+    private fun setupAudioPlayerButton() {
+        //val playAudio = fragView.findViewById<View>(R.id.play_audio_profile_button)
+        //playAudio.setOnClickListener {
+        //playAudio()
+        //Toast.makeText(context, "Ui", Toast.LENGTH_LONG).show()
+        //}
```
Xelowak, line 299: Is it buggy for now ? I suggest you explain why it is commented.

app/src/main/res/layout/item_profile.xml
Xelowak, line 47: Cirrus says that there could be a problem with the API number. Is there an older way to do the same thing ? Otherwise I guess we can change the current min API.

Xelowak (approved):
Nice job overall !
I think that you need to test the constructor using a Parcel and the `writeToParcel` function in `ProfileTest.kt` (if it is possible). Try to test everything you can (I mean that doesn't use Firebase, like the CardStackAdapter), so that you can improve your code coverage, 22% is not really good :'(

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MatchPageFragment.kt
Lollipoke: You're right, it's a bit unusual xD I changed it

app/src/main/java/ch/epfl/sdp/blindly/match/Profile.kt
Lollipoke: It's because readString returns a String? but readInt returns an Int. I guess we could store the variables as String?s but we would have to handle the uncertainty all the way through the activities. I think it's easier to assume the check is done before constructing the profile

app/src/main/res/layout/item_profile.xml
Lollipoke: You're right !

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MatchPageFragment.kt
Lollipoke: Oh I forgot to remove it ^^' I moved it somewhere else

app/src/main/res/layout/item_profile.xml
Lollipoke: I don't think there's an older way to justify Textviews, at least I didn't find any :/ I could remove it but the description text would be aligned to the left. I'm not sure what the best option is atm
I can remove it for now and we can discuss it

app/src/main/java/ch/epfl/sdp/blindly/main_screen/MatchPageFragment.kt
Lollipoke: I'll take a look at it, it's a good idea :D

Lollipoke: Okay so unfortunately it isn't possible, the Parcelable interface requires the parameters to be in the primary constructor

Xelowak: That's sad :(

PR #186, opened 2021-05-06
Refactor RecordingActivity

cmsigrist:
Refactor the RecordignActivity so that it can be called from any Activity

### DIFF
--- a/app/src/androidTest/java/ch/epfl/sdp/blindly/main_screen/ProfilePageTest.kt
+++ b/app/src/androidTest/java/ch/epfl/sdp/blindly/main_screen/ProfilePageTest.kt
@@ -9,7 +9,6 @@ import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.ext.junit.rules.ActivityScenarioRule
 import ch.epfl.sdp.blindly.EditProfile
 import ch.epfl.sdp.blindly.R
-import ch.epfl.sdp.blindly.recording.RecordingActivity
 import ch.epfl.sdp.blindly.settings.Settings
 import ch.epfl.sdp.blindly.user.UserCache
 import ch.epfl.sdp.blindly.user.UserHelper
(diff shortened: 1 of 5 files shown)


Xelowak (approved):
LGTM ! :)

app/src/androidTest/java/ch/epfl/sdp/blindly/main_screen/ProfilePageTest.kt
```diff
@@ -143,29 +144,6 @@ class ProfilePageTest {
...
-
-        onView(withId(R.id.record_button)).perform(click())
-        val fragmentManager = fragment.childFragmentManager
-        val audioFragment = fragmentManager.findFragmentById(R.id.fragment_audio_container_view)
-        assert(audioFragment == null)
-    }
-
-     */
-
     private fun getAudioPlayerFragment(): Fragment? {
```
cgntnr, line 147: Why is this private function hanging out here all alone :D ? It seems other tests are deleted and no one is using this function.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/audio_player/AudioPlayerFragment.kt
```diff
@@ -1,122 +1,118 @@
...
+import ch.epfl.sdp.blindly.user.UserHelper
+import java.io.File
+import javax.inject.Inject
+
+/**
+ * A simple [Fragment] subclass.
+ * Use the [AudioPlayerFragment.newInstance] factory method to
+ * create an instance of this fragment.
+ */
+
```
cgntnr, line 30: I think it's Github's doing but it seems like all lines are removed and added as git diff, which made it hard to find the changes.

cgntnr (approved):
Good amount of code in PR. It's good that some non-null assertions are replaced to avoid problems. The PR changed its purpose midway I believe so I'm not sure @Xelowak 's first review is relavant anymore, because the test added is removed and other changes/refactoring are done after he reviewed.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/audio_player/AudioPlayerFragment.kt
cmsigrist: Yep it's a bit weird sorry :/

PR #207, opened 2021-05-09
add weather service

koko-ng:


Xelowak (approved):
Looks fine to me ! I'm just wondering if we need to support all those languages (it's a lot !).

app/src/main/java/ch/epfl/sdp/blindly/weather/WeatherService.kt
```diff
@@ -0,0 +1,204 @@
...
+            language = DEFAULT_LANG
+        return language
+    }
+    fun nextWeek(
+        loc: BlindlyLatLng,
+        language: String = getLanguage(),
+        callback: WeatherResultCallback
+    ) {
+        val u: HttpUrl? =  url.toHttpUrlOrNull()
+        val url = u!!.newBuilder()
```
cgntnr, line 167: On line 166,  if url is valid ```url``` is converted to http or null as ```u``` , but on line 167 a non-null assertion is put for ```u!!```. It might cause some issues if the url is not valid right?

cgntnr (approved):
Seems good, just added a comment about a non-null assertion.

app/src/main/java/ch/epfl/sdp/blindly/weather/WeatherService.kt
koko-ng: Yes, in fact I use the `toHttpUrlOrNull` as it is the only non-deprecated method to do so, however as the url is a compilation constant it will never be invalid if the tests passed, so as long as we run the tests before publishing anything it will always works.
Nevertheless I added a comment about this.

cgntnr: Great! Thanks for taking the time to do so.

PR #208, opened 2021-05-11
Update user in database and change directories

Xelowak:
I removed the description and added an UID that is not stored in Firebase (with @Exclude).
I managed directories (again) to be clearer for everyone.


r-ismer (approved):
looks fine to me since it is only an addition of a field and package refactor

r-ismer (approved):

PR #209, opened 2021-05-11
Matches view

cgntnr:
Adds a My Matches fragment where you can see your matches, chat with your matches and see them via map. This page is updated instantly via the snapshot of document of a user. So as soon as matches field of a user is adjusted (a match added or removed), we can immediately see them appear/disappear. 

In terms of testing, unfortunately it's still problematic. I tried Mockito and even a fake Activity to attach to and test the adapter. However I had no luck, I'm open to suggestions/helps.


app/src/androidTest/java/ch/epfl/sdp/blindly/main_screen/my_matches/MyMatchesAdapterTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        val viewHolder: RecyclerView.Adapter<MyMatchesAdapter.ViewHolder> =
+            MyMatchesAdapter(list, arrayListOf(), context, listener)
+        MatcherAssert.assertThat(viewHolder.itemCount, IsEqual.equalTo(2))
+    }
+
+    @Test
+    fun itemTypeIsCorrect() {
+        val match1 = MyMatch(NAME_1, UID_1, IS_EXPANDED)
+        val match2 = MyMatch(NAME_2, UID_2, IS_EXPANDED)
+        val context = mock(Context::class.java)
```
cmsigrist, line 41: Just a question, do you need to mock the context? Or could you use ApplicationProvider.getApplicationContext()?

app/src/main/java/ch/epfl/sdp/blindly/main_screen/fragments/MyMatchesFragment.kt
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    /**
+     * Sets up the adapter for recycler view
+     *
+     * @param view Current fragment's view
+     * @param myMatches List of user's matches
+     */
+    private fun setupRecylerView(view: View, myMatches: ArrayList<MyMatch>) {
```
cmsigrist, line 111: I guess this one doesn't need as argument a view: View since fragView is a lateinit var

app/src/main/java/ch/epfl/sdp/blindly/main_screen/fragments/MyMatchesFragment.kt
```diff
@@ -0,0 +1,127 @@
...
+     */
+    @RequiresApi(Build.VERSION_CODES.O)
+    override fun onCreateView(
+        inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?
+    ): View {
+        // Inflate the layout for this fragment
+        fragView = inflater.inflate(R.layout.activity_my_matches, container, false)
+
+        //Needs to be done in a coroutine
+        viewLifecycleOwner.lifecycleScope.launch {
```
cmsigrist, line 86: Feel free to ignore this one, but I think it would make more sense to launch the coroutine in the UserViewModel for example and use it in the fragment. 
For example you'd instantiate the view model here and then in the ini{} of the UserViewModel you could call the getMyMatches()
And since the viewModel can be mocked, you might (?)  be able to increase the coverage for the class (but I'm really not sure)

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/my_matches/MyMatchesAdapter.kt
```diff
@@ -0,0 +1,148 @@
...
+        viewHolder.userNameLayout.setOnClickListener {
+            val notIsExpanded = !my_matches[position].isExpanded
+            toggleLayout(notIsExpanded, viewHolder.expandableChatAndMapLayout)
+            my_matches[position].isExpanded = notIsExpanded
+        }
+
+
+        viewHolder.chatButton.setOnClickListener {
+            val intent = Intent(context, ChatActivity::class.java)
+            val bundle = bundleOf("matchedId" to my_matches[position].uid)
```
cmsigrist, line 111: Maybe create a const for the "matchedId"

cmsigrist (commented):
It looks very good! Thanks for the great work :)
I left a suggestion for testing but I'm reall ynot sure

app/src/main/java/ch/epfl/sdp/blindly/main_screen/fragments/MyMatchesFragment.kt
cgntnr: I'm not sure if it would be possible, in any case I managed to put query of the database in UserRepository so this part should not be where we lose coverage the most. I believe testing the adapter is a better option but I had no luck with multiple tries as mocking some stuff(like Context) resulted in NPEs or I couldn't use a FakeActivity.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/my_matches/MyMatchesAdapter.kt
cgntnr: Yep, will do.

app/src/androidTest/java/ch/epfl/sdp/blindly/main_screen/my_matches/MyMatchesAdapterTest.kt
cgntnr: No not really it was not working anyways, that one is a leftover from other testing tries. I'll leave it out, thanks.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/fragments/MyMatchesFragment.kt
cmsigrist: Yep I see what you mean.
And wouldn't it be possible to test the onClickListener of the Adapter via the MyMatchFragment (I think it's what Melissa and Kilian did for the AudioLibraryAdapter, testing the adapter in the RecordingActivity)

app/src/main/java/ch/epfl/sdp/blindly/main_screen/fragments/MyMatchesFragment.kt
cgntnr: Absolutely right.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/fragments/MyMatchesFragment.kt
cgntnr: Since MyMatchFragment calls the querying database when it's created I can't use it. I actually tried to use RecordingActivity to test MyMatchesAdapter but I had many problems. In the meantime, I managed to add a test with ApplicationProvider.getApplicationContext() and it seemed to work, let's see if it will help the coverage.

cmsigrist: If it's the call to userRepository.getMyMatches0 that poses a problem that's where a viewModel can help, since the latter can be mocked and the fragment could then be instantiated

cgntnr: Yes it might help I believe, though I had some problems before because the matches_list I expected from the query was not returning at the correct time/scope. Maybe we can look at this further together because I'm not sure how I will apply to this usage.

cmsigrist (approved):

app/src/main/res/drawable/ic_baseline_chat.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:autoMirrored="true" android:height="50dp"
+    android:tint="#A786E3" android:viewportHeight="24"
+    android:viewportWidth="24" android:width="50dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="@android:color/white" android:pathData="M20,2L4,2c-1.1,0 -1.99,0.9 -1.99,2L2,22l4,-4h14c1.1,0 2,-0.9 2,-2L22,4c0,-1.1 -0.9,-2 -2,-2zM6,9h12v2L6,11L6,9zM14,14L6,14v-2h8v2zM18,8L6,8L6,6h12v2z"/>
```
koko-ng, line 4: Is it possible to extract the pathData to a file? I suppose it's an icon, it could be good to separate it from the layout

app/src/main/res/drawable/ic_baseline_location.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="50dp" android:tint="#93C6F6"
+    android:viewportHeight="24" android:viewportWidth="24"
+    android:width="50dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="@android:color/white" android:pathData="M12,2C8.13,2 5,5.13 5,9c0,5.25 7,13 7,13s7,-7.75 7,-13c0,-3.87 -3.13,-7 -7,-7zM12,11.5c-1.38,0 -2.5,-1.12 -2.5,-2.5s1.12,-2.5 2.5,-2.5 2.5,1.12 2.5,2.5 -1.12,2.5 -2.5,2.5z"/>
```
koko-ng, line 4: Same as above

app/src/main/res/layout/activity_my_matches.xml
```diff
@@ -0,0 +1,18 @@
...
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="vertical"
+    app:layout_constraintBottom_toBottomOf="parent"
+    app:layout_constraintEnd_toEndOf="parent"
+    app:layout_constraintStart_toStartOf="parent"
+    app:layout_constraintTop_toTopOf="parent">
+    <!--tools:context=".recording.RecordingActivity"> -->
```
koko-ng, line 12: Either update or remove :)

koko-ng (approved):
Overall very good, just the testing that is still a bit low.

app/src/main/res/layout/activity_my_matches.xml
cgntnr: Yep, forgot it there :D

app/src/main/res/drawable/ic_baseline_chat.xml
cgntnr: I actually tried this, to first convert xml file to svg and then from svg to png but I had errors. I don't think it's a big deal so I'll leave this as it is.

PR #225, opened 2021-05-16
Link likes to db

Lollipoke:
- [x] Put the uid of the card into the likes of the current user in the database
- [x] Filter the liked user out of the potential matches
- [x] Handle matches when user like each-other
- [x] Add messages when no profiles are available + when profiles are loading


app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/MatchPageFragment.kt
```diff
@@ -300,4 +324,27 @@ class MatchPageFragment : Fragment(), CardStackListener {
...
+                likedUserId,
+                "matches",
+                otherUser.matches?.plus(currentUserId)
+            )
+            userRepository.updateProfile(
+                currentUserId,
+                "matches",
+                currentUser.matches?.plus(likedUserId)
+            )
+            Toast.makeText(context, "It's a match !", Toast.LENGTH_SHORT).show()
```
cgntnr, line 347: Even the notification is there :D Nice!!

cgntnr (approved):
Nice, clean and concise PR. This should effectively make the app totally functional.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/MatchPageFragment.kt
Lollipoke: Yeah, but it's more a temporary notification xD it's only sent to the user 'provoking' the like so I'm not sure we should keep it haha

cgntnr: Yeah true, the my matches page actually have a listener of matches of Firestore, maybe we can move this there.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/MatchPageFragment.kt
```diff
@@ -117,11 +122,20 @@ class MatchPageFragment : Fragment(), CardStackListener {
...
      *
      * @param direction the direction the card is swiped (left, right)
      */
+    @RequiresApi(Build.VERSION_CODES.N)
     override fun onCardSwiped(direction: Direction) {
+        if (direction == Direction.Right) {
+            likedUserId = currentCardUid
+            val updatedLikesList = currentUser.likes?.plus(likedUserId)
+            viewLifecycleOwner.lifecycleScope.launch {
+                userRepository.updateProfile(currentUserId, "likes", updatedLikesList)
```
cmsigrist, line 135: Maybe use the const in the User class instead: LIKES

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/MatchPageFragment.kt
```diff
@@ -300,4 +324,27 @@ class MatchPageFragment : Fragment(), CardStackListener {
...
+     * too and match them both
+     *
+     */
+    @RequiresApi(Build.VERSION_CODES.N)
+    private suspend fun checkMatch() {
+        val otherUser = userRepository.getUser(likedUserId)
+        if (otherUser?.likes?.contains(currentUserId)!!) {
+            userRepository.updateProfile(
+                likedUserId,
+                "matches",
```
cmsigrist, line 339: Here too, there's also the const MATCHES in the User

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/MatchPageFragment.kt
```diff
@@ -232,16 +247,17 @@ class MatchPageFragment : Fragment(), CardStackListener {
         if (users == null) {
             return listOf()
         }
-        val userid = userHelper.getUserId()!!
-        val currentUser = userRepository.getUser(userid)
+        currentUserId = userHelper.getUserId()!!
+        currentUser = userRepository.getUser(currentUserId)!!
```
cmsigrist, line 251: I think if you were to use a viewModel here instead of calling directly the repository in the UI that you'd be able to maybe test the match fragment a bit more

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/MatchPageFragment.kt
```diff
@@ -300,4 +324,27 @@ class MatchPageFragment : Fragment(), CardStackListener {
...
     }
+
+    /**
+     * When a user likes someone, check if the other liked them
+     * too and match them both
+     *
+     */
+    @RequiresApi(Build.VERSION_CODES.N)
+    private suspend fun checkMatch() {
+        val otherUser = userRepository.getUser(likedUserId)
```
cmsigrist, line 335: Same here if you use a viewModel,  you could then mock it in the tests

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/cards/CardStackAdapter.kt
```diff
@@ -48,10 +49,11 @@ class CardStackAdapter(
         val profile = profiles[position]
         val v = holder.itemView.findViewById(R.id.item_image) as ImageView
         v.setImageResource(R.drawable.background)
-        holder.name_age.text = "${profile.name}, ${profile.age}"
+        holder.nameAge.text = "${profile.name}, ${profile.age}"
```
cmsigrist, line 52: As Cirrus says it would be better to use a placeholder

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/cards/CardStackAdapter.kt
```diff
@@ -48,10 +49,11 @@ class CardStackAdapter(
...
         holder.gender.text = profile.gender
         holder.distance.text = "${profile.distance} km away"
```
cmsigrist, line 54: Same here for the placeholder

cmsigrist (commented):
Good work!
I made some comments that might help to increase the coverage for the match fragment :)

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/MatchPageFragment.kt
Lollipoke: Thank you, I forgot about that !

PR #228, opened 2021-05-18
UI refactor

cmsigrist:
Refactor the UI so that it also fits on bigger screens


app/src/main/res/layout/profile_setup_house_rules.xml
```diff
@@ -9,152 +9,117 @@
...
+
+
+    <TextView
+        android:id="@+id/guideline_4"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_above="@id/guideline_4_text"
+        android:layout_marginStart="@dimen/margin"
+        android:text="@string/u2713_be_proactive"
+        android:textColor="#000000" />
```
koko-ng, line 103: Because we should really tell something in reviews, but isn't #000 the default color anyway? Or maybe the theme has it defined somewhere else? I really won't care that much if you don't take my remark into account, it's not even you that put this line here originally haha.

koko-ng (approved):
Wow thanks, huge work of refactoring the layouts you did here. Congratulation.

app/src/main/res/layout/profile_setup_house_rules.xml
Lollipoke: Actually I think we don't _have to_  write comments if there's nothing to say xD but indeed I think the default colors are changed in the theme xml of the app, to fit our theme

Lollipoke (approved):
Hopefully you can find what's wrong with Cirrus :/ I can't find anything that could make it fail the tests, your refactoring looks clean :D

app/src/main/res/layout/profile_setup_house_rules.xml
cgntnr: Yeah sorry, it was in the beginning of the project. I just wanted something working and something uniform. Not a great concern I think but still we can change it to polish.

cmsigrist: Nothing to be sorry about, it actually turns grey otherwise x)

PR #229, opened 2021-05-19
Annotations refactor

cmsigrist:
Remove all the @RequireAndroid O annotations and most of the Android N (some remain in the audio directory, for the chronometers to work)


app/src/main/java/ch/epfl/sdp/blindly/profile_setup/ProfileBirthday.kt
```diff
@@ -70,12 +71,19 @@ class ProfileBirthday : AppCompatActivity() {
...
-        ).years
+    private fun getAge(year: Int, month: Int, day: Int): Int {
+        val calendar = GregorianCalendar()
+
+        val y = calendar.get(Calendar.YEAR)
+        //For some unknown reason, months are indexed from 0 to 11...
+        val m = calendar.get(Calendar.MONTH) + 1
+        val d = calendar.get(Calendar.DAY_OF_MONTH)
+        var age = y - year
+        if ((m < month) || ((m == month) && (d < day))) {
```
koko-ng, line 82: Maybe comment this lines to explain what they do, it's a bit cryptic to me

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
```diff
@@ -298,17 +298,22 @@ class User private constructor(
...
-            return Period.between(
-                LocalDate.of(
-                    year.toInt(),
-                    month.toInt(),
-                    day.toInt()
-                ),
-                LocalDate.now()
-            ).years
+            val (day, month, year) = birthday.split('.').map { s -> s.toInt() }
+            val calendar = GregorianCalendar()
```
koko-ng, line 303: This snippet is in two different places, maybe a date utils class could be helpful?

koko-ng (commented):
Nice work, thanks!

koko-ng (approved):

.idea/codeStyles/Project.xml
```diff
@@ -1,22 +1,6 @@
...
-      </option>
-      <option name="PACKAGES_IMPORT_LAYOUT">
-        <value>
-          <package name="" alias="false" withSubpackages="true" />
-          <package name="java" alias="false" withSubpackages="true" />
-          <package name="javax" alias="false" withSubpackages="true" />
-          <package name="kotlin" alias="false" withSubpackages="true" />
-          <package name="" alias="true" withSubpackages="true" />
-        </value>
-      </option>
```
Lollipoke, line 19: I think this happens when you don't run connectedCheck, maybe we should keep the previous version ?

app/src/androidTest/java/ch/epfl/sdp/blindly/profile_setup/ProfileBirthdayTest.kt
```diff
@@ -25,7 +25,7 @@ import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
 
-const val TEST_BIRTHDAY = "18.03.2003"
+const val TEST_BIRTHDAY = "16.05.2003"
```
Lollipoke, line 28: What's the point of changing the date ? Both are > 18 no ?

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/cards/CardStackAdapter.kt
```diff
@@ -48,7 +48,7 @@ class CardStackAdapter(
         val profile = profiles[position]
         val v = holder.itemView.findViewById(R.id.item_image) as ImageView
         v.setImageResource(R.drawable.background)
-        holder.name_age.text = "${profile.name}, ${profile.age}"
+        holder.nameAge.text = "${profile.name}, ${profile.age}"
```
Lollipoke, line 51: I fixed that in my PR ;)

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
Lollipoke: I don't know if a class for one method is really necessary, but maybe a public method could be a good idea indeed !

Lollipoke (approved):
Great job !

app/src/androidTest/java/ch/epfl/sdp/blindly/profile_setup/ProfileBirthdayTest.kt
cmsigrist: Yes, indeed, but it was to test the corner case of m == month and d <= day in getUserAge()

.idea/codeStyles/Project.xml
cmsigrist: Ohhhh! That might be it, thanks (it might have come with the update of Android Studio)!

cgntnr: We can simply not add these to the commit anyways :D Would be an easier solution I believe.

cmsigrist: Yep I'll exclude it :D

PR #230, opened 2021-05-19
Fix recordings swipe card

Lollipoke:
- [x] You can play - pause the recordings using the button
- [x] When swiping, the audio of the previous card stops and the new one starts automatically 
- [x] When loading the profiles and when there's no profiles anymore, the play-pause button is unclickable
- [ ]  Tests

Note: the audio on the first card does not start automatically, I thought it would be a bit aggressive, the user could not be ready to hear recordings directly when entering the app


cgntnr:
I agree, it shouldn't start automatically. User might not be in an appropriate environment to listen etc. so we better give that choice to user.

app/src/main/res/values/strings.xml
```diff
@@ -169,7 +169,7 @@
     <string name="my_map_button_description">Map Button</string>
     <string name="name_age_text">%1$s, %2$d</string>
     <string name="distance_text">%1$s km away</string>
-    <string name="no_more_swipes">You swiped everyone :(\nCome back later!</string>
```
cgntnr, line 172: Thanks for the removal of ":(" Our users deserve smileys all around :D

app/src/main/res/values/strings.xml
```diff
@@ -169,7 +169,7 @@
...
-    <string name="no_available_swipe">No profile available for you :(\nCome back later!</string>
+    <string name="no_more_swipes">You swiped everyone...\nCome back later!</string>
```
cgntnr, line 172: Maybe you can use ellipsis character as Cirrus suggests.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/cards/CardStackAdapter.kt
```diff
@@ -36,6 +45,11 @@ class CardStackAdapter(
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolder {
         context = parent.context
         val inflater = LayoutInflater.from(context)
+        //Initialize the mediaPlayers' list
+        for (i in 0 until itemCount) {
```
cgntnr, line 49: Would this be a problem in terms of scaling?  Just asking to get more knowledge.

cgntnr (approved):
Nice fix! Not much to add, couple of Cirrus warnings might be handled. Coverage suffers again but not too much this time and I know it gets hard with the database. Hopefully we can get some of the coverage back with the abstraction PR that is on the way.

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/cards/CardStackAdapter.kt
Lollipoke: Yes, it would probably become a problem... But I couldn't find a better way since the recordings don't come in order, I have to store them in the correct position so the array has to be pre-initialized. To fix this I think we should download matches partially every N swipes, so that the loading doesn't take too much time and the mediaPlayer's initialization isn't problematic either... But as it is right now I can't of a better way

r-ismer (approved):
It is very good code you did a great job !!

app/src/main/java/ch/epfl/sdp/blindly/main_screen/match/cards/CardStackAdapter.kt
cgntnr: I see, thanks for the explanation!

PR #247, opened 2021-05-26
Remove a matched user

cgntnr:
Added ability to remove a user from My Matches. The manipulation in UserRepository is pretty similar to @cmsigrist 's code to Delete an Account, we can reuse it there. The user has a remove button besides the match's name. When clicked, a warning message is prompted and if user proceeds, that match is removed from My Matches.

Most of the manipulation is already in UserRepository however for testing the warning message, @Xelowak can address this in his testing task #241 .

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/blindly/database/UserRepository.kt
+++ b/app/src/main/java/ch/epfl/sdp/blindly/database/UserRepository.kt
@@ -54,12 +54,12 @@ class UserRepository @Inject constructor(
      * Removes another liked or matched user from current user.
      *
      * @param field field to remove a User (either from LIKES or MATCHES)
-     * @param userUid current user's UID
-     * @param matchUid matched user's UID
+     * @param userId current user's ID
+     * @param matchId matched user's ID
      */
-    suspend fun removeMatchFromAUser(field: String, userUid: String, matchUid:String) {
-        var updatedList: ArrayList<String>? = null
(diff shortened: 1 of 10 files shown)


app/src/main/java/ch/epfl/sdp/blindly/database/UserRepository.kt
```diff
@@ -50,6 +50,30 @@ class UserRepository @Inject constructor(
...
     }
 
+    /**
+     * Removes another liked or matched user from current user.
+     *
+     * @param field field to remove a User (either from LIKES or MATCHES)
+     * @param userUid current user's UID
+     * @param matchUid matched user's UID
+     */
+    suspend fun removeMatchFromAUser(field: String, userUid: String, matchUid:String) {
```
cmsigrist, line 60: I went through your whole pull request and have not found anything to comment on... x)
So I don't know, maybe for coherence you should rename the argument userUid to userId and matchId I guess....

cmsigrist (approved):
Great work!

app/src/main/java/ch/epfl/sdp/blindly/database/UserRepository.kt
cgntnr: Sure I can do that :D

Xelowak (approved):
LGTM !

PR #248, opened 2021-05-26
Adding tests

Xelowak:
Trying to add tests to improve our coverage.

I've already tested extras in `ChatActivity`, and I wrote more tests for the `MyMatch` data class.
I added tests for My Matches UI too.
Since I didn't find a way to test removes from MyMatch, I added tests in User to try to improve our overall coverage.


.idea/runConfigurations.xml
```diff
@@ -1,3 +1,14 @@
+<<<<<<< HEAD
```
cgntnr, line 1: I think you should remove these merge artifacts.

cgntnr (approved):
Good refactoring and addition of tests. Thanks!

.idea/runConfigurations.xml
Xelowak: True ! It is weird that it didn’t tell me that I left it. I’ll probably remove the modification anyway since I don’t know why it has been modified.

r-ismer (approved):
All good !

PR #250, opened 2021-05-27
Bugfix audio on cards

Lollipoke:
Fix the bug where swiping the card without listening to the audio would crash the app


cmsigrist (approved):
LGTM! Thanks :)

PR #251, opened 2021-05-28
fix google auth and splash screen

### LINKED ISSUE #240: [BUGFIX] Fix Google authentication

koko-ng:
Fix the authentication with google auth by changing the isNewUser function that didn't worked. Also make the opening of the app safer as it won't launch if the profile has not been set.

Closes #240


cgntnr:
It might be better if you check the failing tests before we could review.

cgntnr (approved):
Thanks for the fix and additional mocked modules. Obligatory --> LGTM!

r-ismer (approved):
LGTM

PR #256, opened 2021-05-28
Fix local DB

r-ismer:
A quick fix that prevent the matches to crash the app


Xelowak (approved):
It seems to be ok, we just need to take care of this problem this week (instead of completely removing the localDB usage)

cgntnr (approved):
Should be ok so that everyone else use the app without problems but we should look for some solutions. Hopefully you can have some ideas already as we are running out of time and if not we try to find altogether a solution.

PR #258, opened 2021-06-01
Fix remove match

cgntnr:
Now removing a match works correctly. When the current user (removing user), removes a match from its matches, that user is removed from likes and matches and put to dislikes so that match won't appear again cards. We also remove the remote user's match (current user) but keep the current user's id in remote's likes so that it won't reappear in its cards.

### DIFF
--- a/app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
+++ b/app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
@@ -186,7 +186,7 @@ class UserRepositoryImpl constructor(
      * @param userId current user's ID
      * @param matchId matched user's ID
      */
-    override suspend fun removeMatchFromRemovingUser(
+    override suspend fun removeMatchFromCurrentUser(
         userId: String,
         matchId: String
     ) {
@@ -209,20 +209,20 @@ class UserRepositoryImpl constructor(
 
     /**
      * Removes the current user from removed user's matches
-     * It's kept in likes of remote user so that they wouldn't reappear
+     * It's kept in likes of remote user so that they don't reappear
      * in their cards
      *
-     * @param removingUserId
+     * @param currentUserId
      * @param removedUserId
      */
     override suspend fun removeCurrentUserFromRemovedMatch(
-        removingUserId: String,
+        currentUserId: String,
         removedUserId: String
     ) {
         val user = getUser(removedUserId)
         if (user != null) {
             var updatedMatchesList = user.matches as ArrayList<String>?
-            updatedMatchesList?.remove(removingUserId)
+            updatedMatchesList?.remove(currentUserId)
             user.uid?.let { updateProfile(it, MATCHES, updatedMatchesList) }
         }
     }
--- a/app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
+++ b/app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
@@ -181,24 +181,49 @@ class UserRepositoryImpl constructor(
 
 
     /**
-     * Removes another liked or matched user from current user.
+     * Removes a matched user from likes and matches and adds it to dislikes.
      *
-     * @param field field to remove a User (either from LIKES or MATCHES)
      * @param userId current user's ID
      * @param matchId matched user's ID
      */
-    override suspend fun removeMatchFromAUser(field: String, userId: String, matchId: String) {
-        var updatedList: ArrayList<String>? = arrayListOf()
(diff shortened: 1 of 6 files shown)


app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
```diff
@@ -181,24 +181,49 @@ class UserRepositoryImpl constructor(
...
-     * Removes another liked or matched user from current user.
+     * Removes a matched user from likes and matches and adds it to dislikes.
      *
-     * @param field field to remove a User (either from LIKES or MATCHES)
      * @param userId current user's ID
      * @param matchId matched user's ID
      */
-    override suspend fun removeMatchFromAUser(field: String, userId: String, matchId: String) {
-        var updatedList: ArrayList<String>? = arrayListOf()
+    override suspend fun removeMatchFromRemovingUser(
```
cmsigrist, line 189: I think the previous name of the function was clearer, I'm not sure what a RemovingUser is... :O

app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
```diff
@@ -181,24 +181,49 @@ class UserRepositoryImpl constructor(
...
+    /**
+     * Removes the current user from removed user's matches
+     * It's kept in likes of remote user so that they wouldn't reappear
+     * in their cards
+     *
+     * @param removingUserId
+     * @param removedUserId
+     */
+    override suspend fun removeCurrentUserFromRemovedMatch(
+        removingUserId: String,
```
cmsigrist, line 219: Same here this removingUser sounds a bit weird x)

app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
```diff
@@ -181,24 +181,49 @@ class UserRepositoryImpl constructor(
...
+            user.uid?.let { updateProfile(it, LIKES, updatedLikesList) }
+            user.uid?.let { updateProfile(it, DISLIKES, updatedDislikesList) }
+            user.uid?.let { updateProfile(it, MATCHES, updatedMatchesList) }
+        }
+    }
+
+
+    /**
+     * Removes the current user from removed user's matches
+     * It's kept in likes of remote user so that they wouldn't reappear
```
cmsigrist, line 212:
```suggestion
     * It's kept in likes of remote user so that they don't reappear
```

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
```diff
@@ -363,6 +378,13 @@ data class User private constructor(
                     }
                     user.likes = newLikes
                 }
+                DISLIKES -> {
```
cmsigrist, line 381: I don't think I missed them, but I think you forgot to add this in the UserUnitTest

cmsigrist (approved):
Looks good to me! 
I suggested some name changes in my review and also to add unit tests for the User since you added a field :D

app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
cgntnr: Yeah since we remove match from the remote user as well, I tried to make it more explicit but I guess it was confusing/ I will rename it to removeMatchFromCurrentUser which should be clearer.

app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
cgntnr: :D Sure why not

app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
cmsigrist: Perfect, thanks! :D

app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
cmsigrist: Don't know kinda souded strange xD (and I neede to find things to review x) )

app/src/main/java/ch/epfl/sdp/blindly/user/User.kt
cgntnr: Yes you are right! I've added them.

r-ismer (approved):
nice work ! looks good to me

PR #259, opened 2021-06-02
Fix audio

cmsigrist:
The recording can now be played offline (if the user logged in at least once while having the internet) 

- [x] If not available locally, retrieve the recording from Firebase storage and save it locally
- [x] Else use the local recording from the app directory
- [x] Change the name of the Audio File that were created during the recording to Recording
- [ ] Add test for the AudioPlayerFragment


.idea/runConfigurations.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cgntnr, line 1: Ah, return of the runConfigurations !

app/src/androidTest/java/ch/epfl/sdp/blindly/splash_screen/SplashScreenActivityTest.kt
```diff
@@ -48,6 +51,15 @@ class SplashScreenActivityTest {
         release()
     }
 
+    @Test
+    fun splashScreenGoesToMainScreen() {
+        Mockito.`when`(user.isLoggedIn()).thenReturn(true)
```
cgntnr, line 56: I think Tibor included this as default behavior in #251, if I'm not mistaken. Might shorten the test but not a big deal, you might as well leave it like this.

cgntnr (approved):
Fix looks fine with some refactoring as well. Though I think maybe we could merge this last, to guarantee we are above 80% coverage, once some additional tests kick in.

app/src/androidTest/java/ch/epfl/sdp/blindly/splash_screen/SplashScreenActivityTest.kt
koko-ng: I think we can leave this as-is. It will be easier to merge and it only isolate this test a bit, which I don't think is a problem.

koko-ng (approved):
Thanks it looks all good to me. I have seen that your renamed Recording to FirebaseRecordings, while I see no problem in doing this maybe I can just explain why I named it that way earlier: My idea is that we abstract everything from Firebase in order to be able to potentially easily switch provider, so we have Recordings which is the interface and FirebaseRecordings whichi is the implementation specifically for Firebase. However I agree that the name Recordings is too vague for the interface.
But anyway the way you did it is also perfectly fine!

PR #261, opened 2021-06-02
Offline mode

r-ismer:
added connection test
only put the user in local db
only dispaly cards if internet is available


app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
```diff
@@ -53,7 +54,7 @@ class UserRepositoryImpl constructor(
             Log.d(TAG, "Found user with uid: $uid in cache")
             return cached
         }
-        /*
```
koko-ng, line 56: Is having a local db and a user cache really useful? If we are too short in time to resolve this, it is ok for me too keep it as is.

koko-ng (approved):
Nice, thanks :)

app/src/main/java/ch/epfl/sdp/blindly/database/UserRepositoryImpl.kt
r-ismer: forgot to uncomment this

r-ismer: hum this is not commented on my side seem like this is not the last version that you are reviewing

r-ismer: yes it is necessary because the user is persistently stored and the cache store every user that is retreived from firestore

.idea/runConfigurations.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="RunConfigurationProducerService">
+    <option name="ignoredProducers">
+      <set>
+        <option value="com.android.tools.idea.compose.preview.runconfiguration.ComposePreviewRunConfigurationProducer" />
+      </set>
+    </option>
+  </component>
+</project>
```
Xelowak, line 10: I don't think this file has to be changed

app/src/androidTest/java/ch/epfl/sdp/blindly/utils/CheckInternetTest.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    @Before
+    fun setup() {
+        context = InstrumentationRegistry.getInstrumentation().context
+    }
+
+    @Test
+    fun isConnectedIsTrueIfConnected() {
+        val bool: Boolean = CheckInternet.internetIsConnected(context)
+        MatcherAssert.assertThat(bool, Matchers.equalTo(true))
```
Xelowak, line 24: It could be nice to import directly MatcherAssert.assertThat so that is it less redundant (like in other tests)

app/src/main/java/ch/epfl/sdp/blindly/utils/CheckInternet.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import android.content.Context
+import android.net.ConnectivityManager
+
+class CheckInternet {
+
+    companion object {
+        const val TAG = "CheckInternet"
+
+        fun internetIsConnected(context: Context): Boolean {
```
Xelowak, line 11: It might be better to rename this function to "isInternetAvailable" like a boolean value

Xelowak (approved):
Nice work ! I advise you to run a CTRL+ALT+L because I saw some inconsistencies.

PR #262, opened 2021-06-02
Fix for weather servive without connection

cgntnr:
When there's no connection, the weather activity was crashing because the Toast wasn't called from runOnUiThread, that's fixed now. Correctly displays Toast and stops refreshing.


koko-ng (approved):
Nice catch thanks!

r-ismer (approved):
LGTM

PR #267, opened 2021-06-03
Firebase recordings test

koko-ng:
Some tests for the firebase recordings, I tried to have generics at the most possible for the mocked Tasks however the base interface is protected in a firebase package, which is a bit constraining :/


cmsigrist (approved):
LGTM!

cgntnr (approved):
We are now very safe with the coverage and it's always good to test more. Thanks!

PR #269, opened 2021-06-04
Fix login

cmsigrist:


r-ismer (approved):
I BELIEVE THIS IS THE FIXXXXX

cgntnr (approved):
Thanks for this last minute fix, one needs to be able to login huh :D

PR #273, opened 2021-06-06
Fix match

Xelowak:
Fix match so that it updates each time the fragment becomes visible.
Did some cleanup too.


Lollipoke (approved):
Thanks for the refactoring and bugfix :D LGTM

